PR #12, opened 2021-02-10
Improve test discoverability

Xiretza:
I came across some problems when trying to package this for the AUR, these patches should fix those without affecting anything else.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -73,7 +73,7 @@ setuptools_setup(
 	},
 	# download_url="https://github.com/vhdl/pyVHDLModel/tarball/0.1.0",
 
-	packages=setuptools_find_packages(),
+	packages=setuptools_find_packages(exclude=["tests", "tests.*",]),
 	classifiers=[
 		"License :: OSI Approved :: Apache Software License",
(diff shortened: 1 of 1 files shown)


umarcor:
@Xiretza, while adding `__init__.py` might be ok, having an `*.ini` file and not naming test files `test_*` is a design decision from Patrick. Is there any chance of making it work in AUR without modifying the file name?

Xiretza:
Apparently `pytest` reads the `pytest.ini` in directories explicitly passed on the commandline, so while not ideal, `pytest tests/` works.

umarcor:
Nice and thanks! LGTM. However, I'll let @Paebbels merge.

Paebbels:
@Xiretza I merged it. Thanks.

Do you need this package as a released version on PyPI or is the modified source ok (via this merge)?
The (full) release pipeline runs only on tags.

Xiretza:
No, Arch packages are built directly from the sources (either release tarballs or, as in this case, the HEAD of the git repo), not pypi releases. Thanks for the merge!

PR #23, opened 2021-07-14
readme: move example to 'example.py' and update it 

umarcor:
The minimal example in the README is slightly outdated. In this PR, I fixed the fields that changed in the latest updates. By the way, I moved it to a separated file, so that users can test it more easily, and we can also easily know when it is outdated again.

### DIFF
--- a/.github/workflows/Pipeline.yml
+++ b/.github/workflows/Pipeline.yml
@@ -222,9 +222,62 @@ jobs:
         run: |
           twine upload dist/*
 
+  VerifyDocs:
+    name: 📓 Verify that example(s) is/are runnable
+    runs-on: ubuntu-latest
+    steps:
+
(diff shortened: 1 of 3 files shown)


Paebbels:
Thanks for updating to the latest API.

Your editor is not obeying to the settings in `.editorconfig` for markdown files. It removed forced linebreaks.  
While the markdown parser used by GitHub is relaxed (and not markdown conformant), other parsers like in PyPI (README is handed over via `setup.py`) should still get a good formatted markdown file.

---------------
I would not move the example from main README as we should demonstrate an easy and short example how it can be used. I agree on having more complex examples elsewhere (mainly as ReST in the main docs).

Paebbels:
Thanks, works like a charm!

PR #29, opened 2021-08-27
v0.13.0

Paebbels:
# New Features
* Moved enumeration `VHDLVersion` from [ProjectModel](https://github.com/edaa-org/pyEDAA.ProjectModel) (former part of [pyIPCMI](https://github.com/Paebbels/pyIPCMI)) to pyVHDLModel.

# Changes
* Updated Sphinx dependencyto 4.2.0

# Bug Fixes
* None

### DIFF
--- a/.github/workflows/Pipeline.yml
+++ b/.github/workflows/Pipeline.yml
@@ -297,12 +297,12 @@ jobs:
           RUN apk add -U --no-cache graphviz
           EOF
 
-      - name: 🛳️ Build documentation from './pyVHDLModel/doc'
+      - name: 🛳️ Build documentation from './pyVHDLModel/doc' and publish to GitHub Pages
         uses: buildthedocs/btd@v0
         with:
           token: ${{ github.token }}
 
-      - name: 📤 Upload artifacts to GitHub Pages
+      - name: 📤 Upload artifacts
         uses: actions/upload-artifact@master
         with:
           name: doc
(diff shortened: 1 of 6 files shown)


.github/workflows/Pipeline.yml
```diff
@@ -297,12 +297,12 @@ jobs:
           RUN apk add -U --no-cache graphviz
           EOF
 
-      - name: 🛳️ Build documentation from './pyVHDLModel/doc'
+      - name: 🛳️ Build documentation from './pyVHDLModel/doc' and publish to GitHub Pages
```
Paebbels, line 300: @umarcor is this change correct?

Why do we have a step to push an artifact?

umarcor: Yes, it is correct. The BTD Action builds the site and pushes it to branch gh-pages, all at once.

Unlike GitLab, the website does not need to be uploaded as an artifact in order to be published. Therefore, the step for pushing the artifact in this case is unrelated to publishing the site. We upload the artifacts for debugging purposes. In each CI run (either push or PR), the artifact remains as a zipfile (see, for instance https://github.com/VHDL/pyVHDLModel/actions/runs/1250153143). So, while the main site hosts the "latest" version only, artifacts allow contributors/maintainers to download some specific variant/build.

Paebbels: Ok, so I mixed it with GitLab.

PR #32, opened 2021-09-27
Add refs to IEEE-P1076.gitlab.io through intersphinx

umarcor:
I'm trying to add references to IEEE-P1076.gitlab.io through intersphinx, with no success. This PR is for showcasing the issue, in the hope that someone can help me understand what's going on.

This is the output I get:

image

- Using role `label`, the targets are not correct.
- Using `ref` without the `VASG` namespace is correct, however, it does not work if the namespace is specified.
- Using `doc` works regardless of the namespace being used.

My understanding is that this might be a bug in Sphinx/Intersphinx.

### DIFF
--- a/doc/Glossary.rst
+++ b/doc/Glossary.rst
@@ -12,3 +12,5 @@ Glossary
      * `1076-2000 <https://ieeexplore.ieee.org/document/893288>`__.
      * `1076-1993 <https://ieeexplore.ieee.org/document/392561>`__.
      * `1076-1987 <https://ieeexplore.ieee.org/document/26487>`__.
+
+     See :ref:`VASG:About:Standards` for further details.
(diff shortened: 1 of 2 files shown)


Paebbels:
> My understanding is that this might be a bug in Sphinx/Intersphinx.

@umarcor are you going to file a bug report?

We're using latest Sphinx v4.2.x, right?

umarcor:
> @umarcor are you going to file a bug report?
> 
> We're using latest Sphinx v4.2.x, right?

@Paebbels, yes we are using latex Sphinx: https://github.com/VHDL/pyVHDLModel/runs/3715512683?check_suite_focus=true#step:5:173.

Before opening a bug report, I would like to understand why references to `python` or `OSVB` namespaces do work, but not for pyVHDLModel or pySystemVerilogModel.

umarcor:
As explained in https://github.com/sphinx-doc/sphinx/issues/9682#issuecomment-928260604, the intersphinx mapping identifier needs to be lowercase. I fixed this PR, and it is now ready to merge.

PR #34, opened 2021-11-07
update logo and favicon

umarcor:


Paebbels:
I think the logo should be above the shields in the README like it's presented in the documentation.
Especially as it acts like a headline.

image

PR #35, opened 2021-11-07
move logo/banner at the top and cleanup shields/badges

umarcor:
See https://github.com/VHDL/pyVHDLModel/pull/34#issuecomment-962591170. At the same time, the shields/badges in the README and the index of the docs are cleaned up to better match ProjectModel.


Paebbels:
I think we can make this PR "read for review" right?

LGTM!

umarcor:
It's ready now (see https://github.com/edaa-org/pySVModel/pull/13#issuecomment-962647876).

PR #37, opened 2021-11-28
shields: use longCache and style flat-square, reorder

umarcor:
As discussed in edaa-org/pySVModel#14, this PR updates the shields in the README and in the index of the site.


umarcor:
@Paebbels, similarly to the ProjectModel, this is also ready to merge.

PR #84, opened 2024-09-26
Add unary reduction expressions

Tcenova:
# New Features
 
* Add unary reduction expressions

### DIFF
--- a/pyVHDLModel/Expression.py
+++ b/pyVHDLModel/Expression.py
@@ -258,6 +258,36 @@ class InverseExpression(UnaryExpression):
 
 
 @export
+class UnaryAndExpression(UnaryExpression):
+	_FORMAT = ("and ", "")
+
+
+@export
+class UnaryNandExpression(UnaryExpression):
+	_FORMAT = ("nand ", "")
+
+
+@export
+class UnaryOrExpression(UnaryExpression):
+	_FORMAT = ("or ", "")
+
+
+@export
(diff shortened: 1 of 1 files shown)


Paebbels:
Please provide these new operators after `class InverseExpression(UnaryExpression)` and before `class AbsoluteExpression(UnaryExpression):` to keep unary operators close together.

I would prefer the operator names `Unary***Expression` instead of `Reduction***Expression`.

Tcenova:
Ill make those changes shortly. As for the naming, I matched the naming of the GHDL node types. Does it matter if they match?

https://github.com/ghdl/ghdl/blob/92a4fc97fb82dff2374c774c3fd6f47c6307e1c6/pyGHDL/libghdl/vhdl/nodes.py#L218-L223

I should have a PR for GHDL soon which uses these changes to add support for translating unary reduction operators in the DOM.

Paebbels:
The naming is independent from GHDL, but it could correlate. GHDL follows the naming of the VHDL standard, but to be honest, the VHDL standard doesn't follow the common naming as used in computer science.

Technically, there are unary, binary and ternary operators. When adding semantics to an operator, it can be an adding, multiplying, or "reducing" operator / operation.

The VHDL LRM uses at many places name for languages entities, that don't match how it's called everywhere else :).

-----
Should I collect further potential changes for pyVHDLModel or should I create a new release, so you can reference it?

Issue #97, opened 2025-06-05
Feature request: port groups


amb5l:
My application needs an object hierarchy to represent VHDL design units and their contents. I want to capture port grouping information - where sets of port clauses have empty lines and/or comments between them - so I have created a fork of pyVHDLModel with support for port groups. It is backward compatible with the standard (flat) way of accessing port information.

Would it make any sense to make a PR to allow port groups to be considered here?

Paebbels:
I checked your addition. It looks quite nice. I think there are just a few modification needed to accept it upstream.

I see you modifications are on the main branch. I suggest to rename you branch to e.g. `amb5l/main` as a so-called modification branch. This allows you to create a pull-request and later update your main with the merged result.

Please create the PR against branch `dev`.

Having a PR allows me to fully check all modifications line-by-line and give you a line-by-line feedback + code suggestions.

amb5l:
> I see the parser is based on ANTLR4. Have you noticed that this implementation is awfully slow? That's why I aborted working on ANTLR4. Please note, ANTLR4 doesn't create a FSM for parsing a language, it learns and trains it's statistical models on the fly creating a huge overhead.
> 
> It's still in discussion to try out ANTRL3, which doesn't do creepy predictions like ANTLR4.
> 
> /cc [@skoehler](https://github.com/skoehler)

Thanks for this insight - very interesting. I have no experience in working with parsers and would never have known.

I have not noticed a performance issue but have only been parsing small test sources. My diagramming app only needs to read entity and component declarations (to create blocks) so I may not have performance problems. Creating a complete diagram from source (code to graphics) is - currently - outside the scope of my project.

I will look at ANTLR3.

Paebbels:
Unfortunately the branch of #99 was deleted or so, so GitHub closed the PR. I was able to reconstruct that code. I'll integrate it in v0.35.0 with some modifications.

@amb5l I'll reference you in the commits using `Co-authored-by:`.

PR #99, opened 2025-06-14
Groups support for entity and component ports, and architecture signal declarations

amb5l:
As discussed in [issue 97](https://github.com/VHDL/pyVHDLModel/issues/97).

### DIFF
--- a/pyVHDLModel/Base.py
+++ b/pyVHDLModel/Base.py
@@ -35,7 +35,7 @@ This module contains parts of an abstract document language model for VHDL.
 Base-classes for the VHDL language model.
 """
 from enum                  import unique, Enum
-from typing                import Type, Tuple, Iterable, Optional as Nullable, Union, cast
+from typing                import Type, Tuple, Iterable, Optional as Nullable, Union, cast, TypeVar, Generic, Dict, List
 
 from pyTooling.Decorators  import export, readonly
 from pyTooling.MetaClasses import ExtendedType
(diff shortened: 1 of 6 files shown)


Paebbels:
Thanks for providing the PR, so I can review the proposed changes. So far I can see the solution is completely different from the first proposal.

I see some drawbacks... give me some time to write a detailed explanation and justification.

So far, I'm missing essential features of a `PortGroup`:
* Name
* Description / Documentation
* Parent reference


------
Other findings:
* A component is not a design unit, thus the mixin name needs to be different.  
  The `TODO` in my code mentions 2 separate mixins for ports and generics, so it could be reused. Solving this TODO is not required for the port groups feature.
* The proposed implementation covers ports and signals, but what about generics, parameters, variables, constants and files.  
  Why should we care for signal groups?  
  Isn't it more important what objects (constant, variable, signal file) are grouped?  
  I see a meaning in groups for interfaces, but not for objects as these are interleaved e.g. also with types and subprograms.

amb5l:
> Thanks for providing the PR, so I can review the proposed changes. So far I can see the solution is completely different from the first proposal.

The current approach results from further study of your codebase but I'm happy to change it as much as needed to align with your preferences.

> I see some drawbacks... give me some time to write a detailed explanation and justification.
> 
> So far, I'm missing essential features of a `PortGroup`:
> 
> * Name
> * Description / Documentation
> * Parent reference

When parsing source code and creating pyVHDLModel objects, I expect groups to be inferred from source proximity, and the name to extracted from comments, or meta-comments. The name is the key of the PortGroups dict. I can see that it may make sense for a group to have a description, in which case we need a "PortGroup" class - subclassed from a "Group" class - a list with a description.

I assume that since the elements of the list will be references to the (flat) ports/signals/etc objects, and those objects have a parent, there is no need to have a parent attribute for a group, or a dict of groups? But it would be easy to add this.

> Other findings:
> 
> * A component is not a design unit, thus the mixin name needs to be different.

No problem. Do you have a suggestion?

>   The `TODO` in my code mentions 2 separate mixins for ports and generics, so it could be reused. Solving this TODO is not required for the port groups feature.

I'm happy to change this as required.

> * The proposed implementation covers ports and signals, but what about generics, parameters, variables, constants and files.

I expect that the grouping functionality will expand to include these, if we get the foundations right.

>   Why should we care for signal groups?

My code often has groups of signals associated with different areas of functionality in the architecture body. For example, different clock domains, or different pipeline stages.

>   Isn't it more important what objects (constant, variable, signal file) are grouped?

If so then let's create grouping functionality for these also.

>   I see a meaning in groups for interfaces, but not for objects as these are interleaved e.g. also with types and subprograms.

See my comment about signal groups above.

Let me know how you would like me to proceed?

Paebbels:
Argh, I see, GitHub deleted `dev` due to branch auto cleanups.

PR #107, opened 2026-04-20
Few patch to improve Vivado code case support

JackCarterSmith:
Hi,

I've used pyVHDLModel parser through pyGHDL.dom to parse some Xilinx Vivado VHDL source.

I've encountered several issues that cause the parser to hang in the following cases:
- subfolder libs structure "use lib.all;" declaration type. Can't be sure if it's allowed in VHDL language, but Vivado use it to directly list all the subfolders declared entity as part of the lib.
pyVHDLModel crash on that because the libraryName value can't be extracted due to the absence of the "package.lib.all" syntax.
- When Vivado use blackboxed IP, it use component to instanciate them. But pyVHDLModel doesn't link component and the corresponding entity properly (or at least I didn't see if and how it does it), and crash.
- When multiple top is present (like when the entities aren't match with correct component and produce 2 or more tree), the exception message does not parse the “roots” values correctly.

So I added a few checks to prevent crashes and exceptions. Ideally, though, I would have preferred that the components be properly linked or treated as black boxes if no entity is found during parsing.

In my humble opinion, this is just “hacks”.
If you have any ideas for improvements or better ways to approach things, I'm all for it.

Thanks again, though, for all the impressive work you've done so far!


Paebbels:
Thanks for your PR.

I'll check it. Some feedback sofar:
* Thanks for the catch on multiple roots.
* The check `if not component.IsBlackbox and component.Entity is not None:` with an additional check for None on the Entity is not correct. It might be a good workaround, but `IsBlackbox` should be True if no entity was found. So the bug/mistake is somewhere on the other side, were `IsBlackbox` isn't populated correctly.  
  I'll check it.
* Do you have an example file name from Vivado were `use lib.all;` appears?  
  Then I'll check it.

Paebbels:
Is your component in an architecture, but not in a package?

Paebbels:
My intermediate state:
* I cherry-picked your change for multiple roots.
* I added linking components if declared in architectures. The change was quite simple:  
  ```diff
  - for package in self.IterateDesignUnits(DesignUnitKind.Package):  # type: Package
  + for designUnit in self.IterateDesignUnits(DesignUnitKind.Package | DesignUnitKind.Architecture):  # type: Union[Package, Architecture]
  ```
* I also found a solution for linking components nested in statements (block statement, generate statement).

--------------
I searched `.../unisims/primitive/*.vhd` for `use\s+\w+\.all;` but I can't find any.

--------------
At next I see this error:
> Unknown expression kind 'Selected_By_All_Name' in expression '2335' at C:\Xilinx\2025.2\data\vhdl\src\unisims\primitive\BUFCE_LEAF.vhd:126:21.

This is caused by `Message.all` in `BUFCE_LEAF.vhd`.

JackCarterSmith:
> Is your component in an architecture, but not in a package?

I'm not sure how Vivado declare its own IP (I see the lib is forced to "xil_defaultlib"), I only see instanciation with component in architecture section + instance in begin section of the architecture.

> My intermediate state:
> 
>     * I cherry-picked your change for multiple roots.
> 
>     * I added linking components if declared in architectures. The change was quite simple:
>       ```diff
>       - for package in self.IterateDesignUnits(DesignUnitKind.Package):  # type: Package
>       + for designUnit in self.IterateDesignUnits(DesignUnitKind.Package | DesignUnitKind.Architecture):  # type: Union[Package, Architecture]
>       ```
> 
>     * I also found a solution for linking components nested in statements (block statement, generate statement).
> 

Thanks for reviewing the fixes, maybe I got trouble with components linking due to generate statement in my VHDL indeed. I'll check it.

> I searched `.../unisims/primitive/*.vhd` for `use\s+\w+\.all;` but I can't find any.
> 

There are in `secureip` directory of `unisims`, "use SECUREIP.all;" at line 35 of `unisims/secureip/bitslice_control.vhd`. Primitives are clean of "use lib.all" statement from what I seen.

> At next I see this error:
> 
> > Unknown expression kind 'Selected_By_All_Name' in expression '2335' at C:\Xilinx\2025.2\data\vhdl\src\unisims\primitive\BUFCE_LEAF.vhd:126:21.
> 
> This is caused by `Message.all` in `BUFCE_LEAF.vhd`.

Yes, I got stuck on it too. (I mentionned it in [this issue](https://github.com/ghdl/ghdl/issues/3222#issuecomment-4205007046) of pyGHDL repo)

PR #126, opened 2026-07-16
Fix crashes in GetAncestor/AllowBlackbox, TopLevel hierarchy check, and PackageInstantiation context items

Paebbels:
## Summary

Four independent, previously-crashing/incorrect behaviors found while building a translation layer on top of `pyGHDL.dom`, plus one code-quality fix. All are covered by new regression tests in `tests/unit/Regression.py`.

### 1. `ModelEntity.GetAncestor()` crashed with an unguarded `AttributeError`

```python
entity = Entity("e")
entity.GetAncestor(Package)
# before: AttributeError: 'NoneType' object has no attribute '_parent'
# after:  VHDLModelException: No ancestor of type 'Package' found for ...
```

The loop walked `parent = parent._parent` with no check for `parent is None`, so reaching the root of the model without a match dereferenced `None._parent`.

### 2. `AllowBlackboxMixin.AllowBlackbox` had the identical unguarded-recursion bug

```python
Entity("e").AllowBlackbox
# before: AttributeError: 'NoneType' object has no attribute 'AllowBlackbox'
# after:  VHDLModelException: AllowBlackbox is not set on ... and no parent is available to inherit it from.
```

Same root cause: if the local value is `None` *and* `Parent` hasn't been assigned yet (a normal transient state for an object built standalone, e.g. in tests), it recurses into `None.AllowBlackbox`.

### 3. `Design.TopLevel` misreported a valid hierarchy as "not yet computed"

`CreateHierarchyGraph()`'s first step unconditionally copies every entity/architecture vertex into `_hierarchyGraph`; edges are only added afterwards where an actual relationship (e.g. `Implementation`, instantiation) exists. So a design containing a single entity with no (yet-analyzed) architecture legitimately produces `VertexCount == 1`, `EdgeCount == 0` - and the old `if self._hierarchyGraph.EdgeCount == 0: raise ...` misclassified this as an uncomputed hierarchy. Switched the check to `VertexCount == 0`, which is the invariant that's actually unconditionally populated by step 1.

### 4. `PackageInstantiation` had no way to carry `contextItems`

`PackageInstantiation.__init__` never accepted/forwarded a `contextItems` parameter to `Package.__init__`, even though every other primary design unit (`Entity`, `Architecture`, `Package`, `PackageBody`, `Configuration`) does. This meant `library`/`use`/`context` clauses preceding a `package ... is new ...;` design unit had no representation at all - not "not yet implemented", just structurally impossible. Consumers (e.g. `pyGHDL.dom`) that compute these clauses had nowhere to put them.

### 5. `Regions.IndexDeclaredItems`: replaced a stray `print()` with the established warning idiom

```python
elif isinstance(item, Variable):
    # FIXME: raise a warning
    print(f"IndexDeclaredItems - {item._identifiers}")
```

Replaced with `WarningCollector.Raise(NotImplementedWarning(...))`, matching the same `pyTooling.Warning`-based idiom already used 9x elsewhere in `pyVHDLModel/__init__.py` (and mirrored by `pyGHDL.dom`'s use of the same `pyTooling.Warning.WarningCollector`).

## Testing

Added `tests/unit/Regression.py`, one positive + one negative test per fix. Full suite: `70 passed` (68 previously existing/added earlier + 2 for the warning fix), no regressions.

## Notes for reviewers

- Item 4 is a prerequisite for a companion fix in `pyGHDL.dom` (`PackageInstantiation.parse()` currently computes `contextItems` for a package instantiation design unit and silently discards them, since there was no constructor parameter to pass them to).
- Not in scope for this PR (flagged for a possible follow-up, happy to open separate issues if useful):
  - `Symbol` classes don't inherit from `ModelEntity`, yet code throughout (`Architecture.__init__`, `PackageInstantiation.__init__`, ...) does `someSymbol.Parent = self` - this currently "works" only because `Symbol`'s metaclass doesn't restrict attributes, creating an undeclared ad-hoc attribute rather than a real `ModelEntity.Parent` property. Already flagged in `Symbol.py` with `# QUESTION: Why is it not a ModelEntity?`.
  - `pyGHDL.dom`'s `Document.translate()` and `DesignUnit.Context.parse()` duplicate the same context-item-parsing loop verbatim; a shared helper would reduce future-divergence risk.


pyVHDLModel/Base.py
```diff
@@ -134,10 +134,24 @@ def Parent(self, parent: 'ModelEntity') -> None:
...
+		:raises VHDLModelException: If the root of the model is reached without finding an ancestor of the requested
+		                            type.
+		"""
 		parent = self._parent
-		while not isinstance(parent, type):
+		while parent is not None and not isinstance(parent, type):
 			parent = parent._parent
 
+		if parent is None:
+			from pyVHDLModel.Exception import VHDLModelException
```
Paebbels: Place imports at beginning of the file clsoe the the other pyVHDLModel imports.

pyVHDLModel/Base.py
```diff
@@ -134,10 +134,24 @@ def Parent(self, parent: 'ModelEntity') -> None:
...
+
+			raise VHDLModelException(f"No ancestor of type '{type.__name__}' found for {self!r}.")
```
Paebbels: Can't you use the loop with `else` construct to handle the error if reaching the end of the chain without finding the requested element?

pyVHDLModel/Base.py
```diff
@@ -134,10 +134,24 @@ def Parent(self, parent: 'ModelEntity') -> None:
 		self._parent = parent
 
 	def GetAncestor(self, type: Type) -> 'ModelEntity':
+		"""
+		Iterate the parent chain - starting at this model entity - upwards (toward the root of the model) and return the
+		closest ancestor whose type matches ``type``.
```
Paebbels: Try to keep your summary in one line (up to 100..120 chars).

Further explanations can be made in a second paragraph (one empty line in between). The summary is shown in lists of methods per class and might get shortend by Sphinx if too long.

pyVHDLModel/Common.py
```diff
@@ -72,10 +72,20 @@ def AllowBlackbox(self) -> bool:
 		1. If allow blackbox property is locally set, return the local value,
 		2. Otherwise, return allow blackbox value from parent object.
 
-		:returns: If blackboxes are allowed.
+		:returns:                   If blackboxes are allowed.
+		:raises VHDLModelException: If neither a local value is set nor a parent object is available to inherit the
+		                            value from (e.g. this object hasn't been added to a design hierarchy yet).
```
Paebbels: Don't give examples here -> not good style.

pyVHDLModel/Common.py
```diff
@@ -72,10 +72,20 @@ def AllowBlackbox(self) -> bool:
...
 		"""
 		if self._allowBlackbox is None:
-			return self._parent.AllowBlackbox
+			parent = getattr(self, "_parent", None)
```
Paebbels: The field `_parent` must always exist. If you identified an element int he hierarchy chain that doesn't implement it, we need to fix it there.

pyVHDLModel/Instantiation.py
```diff
@@ -83,8 +83,15 @@ class PackageInstantiation(Package, GenericInstantiationMixin):  # TODO: maybe a
...
 
-	def __init__(self, identifier: str, genericPackage: PackageReferenceSymbol, documentation: Nullable[str] = None, parent: Nullable[ModelEntity] = None) -> None:
-		super().__init__(identifier, documentation=documentation, parent=parent)
+	def __init__(
+		self,
+		identifier: str,
+		genericPackage: PackageReferenceSymbol,
+		contextItems: Nullable[Iterable[ContextUnion]] = None,
+		documentation: Nullable[str] = None,
+		parent: Nullable[ModelEntity] = None
```
Paebbels:
```suggestion
		self,
		identifier:     str,
		genericPackage: PackageReferenceSymbol,
		contextItems:   Nullable[Iterable[ContextUnion]] = None,
		documentation:  Nullable[str] =                    None,
		parent:         Nullable[ModelEntity] =            None
```

pyVHDLModel/Instantiation.py
```diff
@@ -83,8 +83,15 @@ class PackageInstantiation(Package, GenericInstantiationMixin):  # TODO: maybe a
...
+	) -> None:
+		super().__init__(identifier, contextItems=contextItems, documentation=documentation, parent=parent)
```
Paebbels:
```suggestion
		super().__init__(identifier, contextItems, documentation=documentation, parent=parent)
```

pyVHDLModel/Regions.py
```diff
@@ -186,8 +188,11 @@ def IndexDeclaredItems(self) -> None:
 					self._signals[normalizedIdentifier] = item
 					self._namespace._elements[normalizedIdentifier] = item
 			elif isinstance(item, Variable):
-				# FIXME: raise a warning
-				print(f"IndexDeclaredItems - {item._identifiers}")
+				# TODO: variables declared in a concurrent declaration region (e.g. shared variables outside a
+				#       protected type) are not yet indexed into a dedicated namespace/lookup table.
+				WarningCollector.Raise(
+					NotImplementedWarning(f"IndexDeclaredItems: variable(s) {item._identifiers} are not yet indexed.")
```
Paebbels: list identifiers in single-quotes and as coma separated list of multiple. You can use `", ".join(f"'{i}'" for i in item.identifiers)` within the outer f-string. Maybe you need to fix some my quick code.

tests/unit/Hierarchy.py
Paebbels, line 1: I think most tests in here are checking instantiations. please extend `Instantiate.py`. Review this file to understand it's structure I started and continue with my intents.

The test in `GetAncestor` and `AllowBlackbox` span multiple classes/levels in the model. Maybe use a `Hierarchy.py` module and collect tests traversing the hierarchy. The class name should then express the group of checks e.g. your AllowBlackBox tests.

pyVHDLModel/Base.py
Paebbels: Tried this - it creates a genuine circular import: `Base -> Exception -> Symbol -> Base` (Exception.py imports Symbol.py, which imports `Range` from Base.py). Reproduced:

```
ImportError: cannot import name 'VHDLModelException' from partially initialized module 'pyVHDLModel.Exception' (most likely due to a circular import)
```

Kept the deferred import inside the method with a comment explaining why. Happy to revisit if you'd rather restructure `Exception.py` to not depend on `Symbol.py` (could split `VHDLModelException` into its own dependency-free module), but that felt out of scope for this PR - let me know if you'd like it as a follow-up.

pyVHDLModel/Base.py
Paebbels: Done - rewrote as while/else.

pyVHDLModel/Base.py
Paebbels: Done - shortened to one line, moved details to a second paragraph.

pyVHDLModel/Common.py
Paebbels: Done - removed the example.

pyVHDLModel/Common.py
Paebbels: Confirmed: verified all 9 use sites of AllowBlackboxMixin (Design, Library, Entity, Architecture, Package, Component, ConcurrentBlockStatement, GenerateBranch, GenerateStatement, ConcurrentCase) and every one inherits ModelEntity directly or transitively. Dropped the getattr and use self._parent directly, only guarding for None.

pyVHDLModel/Instantiation.py
Paebbels: Applied as suggested.

pyVHDLModel/Instantiation.py
Paebbels: Applied as suggested.

pyVHDLModel/Regions.py
Paebbels: Done, using the public Identifiers property instead of the private tuple: ', '.join(f"'{i}'" for i in item.Identifiers)

tests/unit/Hierarchy.py
Paebbels: Done - removed Regression.py and redistributed: PackageInstantiation contextItems tests into Instantiate.py/SimpleInstance (matching test_Package/test_PackageBody), GetAncestor/AllowBlackbox into a new Hierarchy.py (parent-chain traversal spanning multiple classes), and TopLevel/IndexDeclaredItems(Variable-warning) into Analyze.py alongside CreateHierarchyGraph/IndexArchitectures which they exercise.

pyVHDLModel/Regions.py
```diff
@@ -186,8 +188,10 @@ def IndexDeclaredItems(self) -> None:
 					self._signals[normalizedIdentifier] = item
 					self._namespace._elements[normalizedIdentifier] = item
 			elif isinstance(item, Variable):
-				# FIXME: raise a warning
-				print(f"IndexDeclaredItems - {item._identifiers}")
+				# TODO: variables declared in a concurrent declaration region (e.g. shared variables outside a
+				#       protected type) are not yet indexed into a dedicated namespace/lookup table.
+				identifiers = ", ".join(f"'{i}'" for i in item.Identifiers)
```
Paebbels:
```suggestion
				identifiers = ", ".join(f"'{i}'" for i in item._identifiers)
```

pyVHDLModel/Common.py
```diff
@@ -72,9 +72,16 @@ def AllowBlackbox(self) -> bool:
...
 		"""
 		if self._allowBlackbox is None:
+			if self._parent is None:
+				from pyVHDLModel.Exception import VHDLModelException
+
+				raise VHDLModelException(f"AllowBlackbox is not set on {self!r} and no parent is available to inherit it from.")
+
 			return self._parent.AllowBlackbox
 		else:
 			return self._allowBlackbox
```
Paebbels:
```suggestion
		if self._allowBlackbox is not None:
			return self._allowBlackbox
		elif self._parent is None:
			from pyVHDLModel.Exception import VHDLModelException

			raise VHDLModelException(f"AllowBlackbox is not set on {self!r} and no parent is available to inherit it from.")
		else:
			return self._parent.AllowBlackbox
```

tests/unit/Hierarchy.py
```diff
@@ -0,0 +1,95 @@
...
+#  |_|    |___/                                                                                                        #
+# ==================================================================================================================== #
+# Authors:                                                                                                             #
+#   Patrick Lehmann                                                                                                    #
+#                                                                                                                      #
+# License:                                                                                                             #
+# ==================================================================================================================== #
+# Copyright 2017-2026 Patrick Lehmann - Boetzingen, Germany                                                            #
+# Copyright 2016-2017 Patrick Lehmann - Dresden, Germany                                                               #
```
Paebbels:
```suggestion
# Copyright 2026-2026 Patrick Lehmann - Boetzingen, Germany                                                            #
```

tests/unit/Hierarchy.py
```diff
@@ -0,0 +1,95 @@
...
+# SPDX-License-Identifier: Apache-2.0                                                                                  #
+# ==================================================================================================================== #
+#
+"""Tests for traversing the model's parent-chain hierarchy, spanning multiple classes/levels of the model."""
+from unittest import TestCase
+
+from pyVHDLModel             import Design, Library, VHDLModelException
+from pyVHDLModel.DesignUnit   import Entity, Architecture, Package
+from pyVHDLModel.Symbol       import EntitySymbol
+from pyVHDLModel.Name         import SimpleName
```
Paebbels:
```suggestion
from pyVHDLModel             import Design, Library, VHDLModelException
from pyVHDLModel.DesignUnit  import Entity, Architecture, Package
from pyVHDLModel.Symbol      import EntitySymbol
from pyVHDLModel.Name        import SimpleName
```

PR #127, opened 2026-07-18
PackageInstantiation: accept and forward genericAssociations

Paebbels:
## Summary

Follow-up to #126 (merged). Adds a `genericAssociations` parameter to `PackageInstantiation.__init__`, matching the pattern already used by `ComponentInstantiation`/`EntityInstantiation` in `Concurrent.py`.

Previously `PackageInstantiation.__init__` always hardcoded `self._genericAssociations = []` with no constructor parameter at all, so there was no way to populate it - even a caller that had already parsed a `generic map (...)` aspect had nowhere to put it.

## Motivation

A companion change in `pyGHDL.dom` (`PackageInstantiation.parse()`) now reads `Get_Generic_Map_Aspect_Chain` on the package instantiation node (which - unlike `Get_Generic_Chain`/`Get_Uninstantiated_Package_Decl`, which stay `Null_Iir` until semantic analysis resolves and macro-expands the referenced package - **is** populated right after parsing) and needs to forward the resulting associations here.

Verified end-to-end against real GHDL analysis (built from the nightly release, not just unit-tested in isolation) with:
```vhdl
package GenericPackage is
  generic (WIDTH : natural := 8);
  constant MAX_VALUE : natural := 2**WIDTH - 1;
end package;

library ieee;
use ieee.std_logic_1164.all;

package InstantiatedPackage is new work.GenericPackage
  generic map (WIDTH => 16);
```
results in `InstantiatedPackage.GenericAssociations == [GenericAssociationItem(Formal='WIDTH', Actual=IntegerLiteral(16))]`.

## Testing

Extended `tests/unit/Instantiate.py::SimpleInstance::test_PackageInstantiation` and added `test_PackageInstantiation_withoutContextItemsOrGenerics`. Full suite: `73 passed`, no regressions.

### DIFF
--- a/pyVHDLModel/Instantiation.py
+++ b/pyVHDLModel/Instantiation.py
@@ -85,11 +85,12 @@ class PackageInstantiation(Package, GenericInstantiationMixin):  # TODO: maybe a
 
 	def __init__(
 		self,
-		identifier:     str,
-		genericPackage: PackageReferenceSymbol,
-		contextItems:   Nullable[Iterable[ContextUnion]] = None,
-		documentation:  Nullable[str] =                    None,
-		parent:         Nullable[ModelEntity] =            None
+		identifier:          str,
+		genericPackage:      PackageReferenceSymbol,
+		contextItems:        Nullable[Iterable[ContextUnion]] = None,
+		genericAssociations: Nullable[Iterable[GenericAssociationItem]] = None,
+		documentation:       Nullable[str] =                    None,
+		parent:              Nullable[ModelEntity] =            None
 	) -> None:
 		super().__init__(identifier, contextItems, documentation=documentation, parent=parent)
 		GenericEntityInstantiationMixin.__init__(self)
@@ -99,6 +100,9 @@ class PackageInstantiation(Package, GenericInstantiationMixin):  # TODO: maybe a
 
 		# TODO: extract to mixin
 		self._genericAssociations = []
+		if genericAssociations is not None:
+			for association in genericAssociations:
+				self._genericAssociations.append(association)
 
 	@readonly
 	def PackageReference(self) -> PackageReferenceSymbol:
--- a/pyVHDLModel/Instantiation.py
+++ b/pyVHDLModel/Instantiation.py
@@ -87,10 +87,10 @@ class PackageInstantiation(Package, GenericInstantiationMixin):  # TODO: maybe a
 		self,
 		identifier:          str,
 		genericPackage:      PackageReferenceSymbol,
-		contextItems:        Nullable[Iterable[ContextUnion]] = None,
+		contextItems:        Nullable[Iterable[ContextUnion]] =           None,
 		genericAssociations: Nullable[Iterable[GenericAssociationItem]] = None,
-		documentation:       Nullable[str] =                    None,
-		parent:              Nullable[ModelEntity] =            None
+		documentation:       Nullable[str] =                              None,
+		parent:              Nullable[ModelEntity] =                      None
 	) -> None:
 		super().__init__(identifier, contextItems, documentation=documentation, parent=parent)
 		GenericEntityInstantiationMixin.__init__(self)
--- a/tests/unit/Instantiate.py
+++ b/tests/unit/Instantiate.py
@@ -44,11 +44,10 @@ from pyVHDLModel.Symbol import AllPackageMembersReferenceSymbol, ContextReferenc
 from pyVHDLModel.Symbol import ArchitectureSymbol, PackageSymbol, EntityInstantiationSymbol
 from pyVHDLModel.Symbol import ComponentInstantiationSymbol, ConfigurationInstantiationSymbol
 from pyVHDLModel.Expression import IntegerLiteral, FloatingPointLiteral
-from pyVHDLModel.Type import Subtype, IntegerType, RealType, ArrayType, RecordType
-from pyVHDLModel.DesignUnit import Package, PackageBody, Context, Entity, Architecture, Configuration
-from pyVHDLModel.DesignUnit import LibraryClause
-from pyVHDLModel.Association import GenericAssociationItem
-from pyVHDLModel.Expression import IntegerLiteral
+from pyVHDLModel.Type          import Subtype, IntegerType, RealType, ArrayType, RecordType
+from pyVHDLModel.DesignUnit    import Package, PackageBody, Context, Entity, Architecture, Configuration
+from pyVHDLModel.DesignUnit    import LibraryClause
+from pyVHDLModel.Association   import GenericAssociationItem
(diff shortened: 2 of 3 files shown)


pyVHDLModel/Instantiation.py
```diff
@@ -85,11 +85,12 @@ class PackageInstantiation(Package, GenericInstantiationMixin):  # TODO: maybe a
...
-		genericPackage: PackageReferenceSymbol,
-		contextItems:   Nullable[Iterable[ContextUnion]] = None,
-		documentation:  Nullable[str] =                    None,
-		parent:         Nullable[ModelEntity] =            None
+		identifier:          str,
+		genericPackage:      PackageReferenceSymbol,
+		contextItems:        Nullable[Iterable[ContextUnion]] = None,
+		genericAssociations: Nullable[Iterable[GenericAssociationItem]] = None,
+		documentation:       Nullable[str] =                    None,
+		parent:              Nullable[ModelEntity] =            None
```
Paebbels:
```suggestion
		contextItems:        Nullable[Iterable[ContextUnion]] =           None,
		genericAssociations: Nullable[Iterable[GenericAssociationItem]] = None,
		documentation:       Nullable[str] =                              None,
		parent:              Nullable[ModelEntity] =                      None
```

tests/unit/Instantiate.py
```diff
@@ -47,6 +47,8 @@
 from pyVHDLModel.Type import Subtype, IntegerType, RealType, ArrayType, RecordType
 from pyVHDLModel.DesignUnit import Package, PackageBody, Context, Entity, Architecture, Configuration
 from pyVHDLModel.DesignUnit import LibraryClause
+from pyVHDLModel.Association import GenericAssociationItem
+from pyVHDLModel.Expression import IntegerLiteral
 from pyVHDLModel.Instantiation import PackageInstantiation
```
Paebbels:
```suggestion
from pyVHDLModel.Type          import Subtype, IntegerType, RealType, ArrayType, RecordType
from pyVHDLModel.DesignUnit    import Package, PackageBody, Context, Entity, Architecture, Configuration
from pyVHDLModel.DesignUnit    import LibraryClause
from pyVHDLModel.Association   import GenericAssociationItem
from pyVHDLModel.Expression    import IntegerLiteral
from pyVHDLModel.Instantiation import PackageInstantiation
```

tests/unit/Instantiate.py
```diff
@@ -430,19 +432,28 @@ def test_PackageInstantiation(self) -> None:
 				LibraryReferenceSymbol(SimpleName("ieee")),
 			]),
 		]
-		packageInstantiation = PackageInstantiation("pack_inst_1", packageReference, contextItems, parent=None)
+		genericAssociations = [
+			GenericAssociationItem(IntegerLiteral(16), SimpleName("WIDTH")),
```
Paebbels: Why is the order of actual and formal swapped?
We have `WIDTH => 16`, thus the `GenericAssociationItem` should take first the actual and then the format. Please check the definition of `GenericAssociationItem`'s initializer. Maybe it needs to be swapped in dunder init.

tests/unit/Instantiate.py
```diff
@@ -430,19 +432,28 @@ def test_PackageInstantiation(self) -> None:
...
+			"pack_inst_1", packageReference, contextItems, genericAssociations, parent=None
+		)
 
 		self.assertIsNotNone(packageInstantiation)
 		self.assertEqual("pack_inst_1", packageInstantiation.Identifier)
 		self.assertIs(packageReference, packageInstantiation.PackageReference)
 		self.assertEqual(1, len(packageInstantiation.ContextItems))
+		self.assertEqual(1, len(packageInstantiation.GenericAssociations))
+		self.assertEqual(16, packageInstantiation.GenericAssociations[0].Actual.Value)
+		self.assertEqual("WIDTH", packageInstantiation.GenericAssociations[0].Formal.Identifier)
```
Paebbels:
```suggestion
		self.assertEqual("WIDTH", packageInstantiation.GenericAssociations[0].Formal.Identifier)
		self.assertEqual(16, packageInstantiation.GenericAssociations[0].Actual.Value)
```

pyVHDLModel/Instantiation.py
Paebbels: Applied as suggested.

tests/unit/Instantiate.py
Paebbels: Applied as suggested, and also dropped a redundant duplicate 'from pyVHDLModel.Expression import IntegerLiteral' I had accidentally added - IntegerLiteral was already imported earlier in this file alongside FloatingPointLiteral.

tests/unit/Instantiate.py
Paebbels: Good catch to double-check, but it's actually not swapped relative to the current definition - AssociationItem.__init__(self, actual, formal=None) takes actual first, so GenericAssociationItem(IntegerLiteral(16), SimpleName('WIDTH')) is (actual=16, formal=WIDTH), which is the correct pairing for 'WIDTH => 16' (verified: association.Formal.Identifier == 'WIDTH', association.Actual.Value == 16).

The real issue is that positional args don't make that obvious from the call site. I checked whether to swap AssociationItem's parameter order to (formal, actual) instead: it's the exact order already used at every production call site of GenericAssociationItem/PortAssociationItem/ParameterAssociationItem, in pyGHDL.dom's GetMapAspect() (pyGHDL/dom/_Translate.py:731,739 - 'yield cls(item, actual, formal)'), which is exercised for every entity/component instantiation's generic and port maps (pyGHDL/dom/Concurrent.py:84-100). Swapping the base class order would ripple through all of those for no functional benefit.

Used keyword arguments here instead (GenericAssociationItem(actual=..., formal=...)), which fully resolves the readability concern without touching that shared, already-exercised API - and added a comment explaining why.

tests/unit/Instantiate.py
Paebbels: Applied as suggested.

tests/unit/Instantiate.py
Paebbels: You are right, sorry for the workaround - fixed properly instead.

AssociationItem.__init__ now takes (formal, actual), both required positional (dropped formal'''s default, since every real call site already passed both explicitly). Updated the three downstream call sites this touches in pyGHDL.dom (GenericAssociationItem/PortAssociationItem/ParameterAssociationItem in Concurrent.py, and GetMapAspect() in _Translate.py) to match.

Verified against real GHDL analysis (not just unit tests in isolation): the full existing testsuite/pyunit/dom suite still passes, including StopWatch'''s real component instantiations with generic/port maps, which exercise this exact code path - so the reorder didn'''t silently break anything downstream.

PR #128, opened 2026-07-18
Rename genericAssociations/portAssociations/parameterMappings to *AssociationItems for consistency

Paebbels:
## Summary

Follow-up naming consistency pass, triggered by review discussion on #127: the class held in every one of these lists is `GenericAssociationItem`/`PortAssociationItem`/`ParameterAssociationItem`, so the parameter/property name should be the plural of the class name - matching `genericItems` (plural of the class held there, `GenericInterfaceItemMixin`) and mirroring `PackageInstantiation.genericAssociationItems` (introduced in #127).

| Old name | New name | Where |
|---|---|---|
| `genericAssociations` / `GenericAssociations` | `genericAssociationItems` / `GenericAssociationItems` | `Concurrent.py`: `Instantiation`, `ComponentInstantiation`, `EntityInstantiation`, `ConfigurationInstantiation`; `Instantiation.py`: `PackageInstantiation` |
| `portAssociations` / `PortAssociations` | `portAssociationItems` / `PortAssociationItems` | same four `Concurrent.py` classes |
| `parameterMappings` / `ParameterMappings` | `parameterAssociationItems` / `ParameterAssociationItems` | `Common.py`: `ProcedureCallMixin`; `Concurrent.py`: `ConcurrentProcedureCall`; `Sequential.py`: `ProcedureCallStatement` |

`parameterMappings` was the odd one out (different word, not just missing "Items") - I checked whether it might be intentionally different, since VHDL's LRM never calls a procedure call's actual parameter part a "map" (unlike generic/port map aspects, which the LRM does call a "map"), but was asked to rename it too for full cross-codebase consistency with pyGHDL.dom, so it's included here.

This is a breaking rename for any external code constructing these classes directly with keyword arguments or reading `.GenericAssociations`/`.PortAssociations`/`.ParameterMappings` - a companion PR in `pyGHDL.dom` (which is the only consumer I'm aware of) updates all matching call sites: https://github.com/paebbels/GHDL/pull/5

## Testing

Updated `tests/unit/Instantiate.py`. Full suite: `73 passed`, no regressions. Also re-verified end-to-end against real GHDL analysis via the companion pyGHDL.dom change - including a new manual check of `ParameterAssociationItems` on a real procedure-call statement (`log("hello", 3)`), which wasn't covered by any existing fixture.

PR #129, opened 2026-07-20
Fix Function.ReturnType crash and Subprogram genericItems/parameterItems constructor gap

Paebbels:
## Summary

Found while auditing pyGHDL.dom for untranslated/silently-broken constructs. `Function.ReturnType`
crashes on every instance, reproducibly:

```python
>>> from pyVHDLModel.Subprogram import Function
>>> Function("my_func").ReturnType
AttributeError: 'Function' object has no attribute '_returnType'
```

`Function.__init__` declares `_returnType: Subtype` as a type annotation and has a `ReturnType`
readonly property reading `self._returnType` - but never assigns it, and doesn't even accept a
`returnType` parameter.

Same root problem, one level up: `Subprogram.__init__` (shared base of `Function`/`Procedure`)
doesn't accept `genericItems`/`parameterItems`/`declaredItems`/`statements` at all - it
unconditionally sets all four to `[]`.

**How pyGHDL.dom currently copes**: reaches directly into `self._returnType`/`self._genericItems`/
`self._parameterItems` after calling `super().__init__()`, bypassing the constructor - marked
`# TODO: move to model` in that codebase, i.e. already flagged as a workaround for a gap here.

## Changes

- `Subprogram.__init__` now accepts `genericItems`/`parameterItems`/`declaredItems`/`statements`,
  wiring `.Parent` correctly on each item (previously never done at all, since the lists were
  always empty).
- `Function.__init__` now accepts and stores `returnType` (required - every function has one).
- `Procedure`/`ProcedureMethod`/`FunctionMethod` forward the new parameters through.
- `ReturnType`'s type hint was `pyVHDLModel.Type.Subtype`, but every real caller actually passes a
  `Symbol` (a reference to the return type, not an owned type definition) - retyped as
  `SubtypeSymbol` to match actual usage.

One thing worth a second opinion: `Subprogram` reimplements generic/parameter item storage instead
of composing `WithGenericsMixin`/`WithParametersMixin` (which already do this correctly elsewhere,
including `.Parent` wiring). I kept `Subprogram`'s own storage for this fix to keep the change
narrowly scoped to "make the constructor work correctly" rather than also restructuring the class
hierarchy - but composing those mixins instead might be the more consistent long-term shape. Happy
to redo it that way if you'd rather.

## Companion change

pyGHDL.dom's `Function`/`Procedure` now use the fixed constructor instead of the private-field
workaround, and additionally now read `IsPure` from `Get_Pure_Flag` (previously never read at all -
every function silently defaulted to `IsPure=True` regardless of `impure`). Verified end-to-end
against real GHDL analysis:

```vhdl
function double(x : integer) return integer is ... end function;
impure function get_random return integer is ... end function;
```
→ `double.IsPure == True`, `get_random.IsPure == False`, both `.ReturnType` populated correctly
(previously would have crashed on `.ReturnType` access).

## Testing

Full suite: `73 passed`, no regressions. This branch is stacked on `naming/genericAssociationItems`
(#128, still open) since I'm actively working across both, to avoid reintroducing the old
`genericAssociations` naming - please merge/rebase order as you see fit.

### DIFF
--- a/pyVHDLModel/Instantiation.py
+++ b/pyVHDLModel/Instantiation.py
@@ -80,17 +80,17 @@ class FunctionInstantiation(Function, SubprogramInstantiationMixin):
 
 @export
 class PackageInstantiation(Package, GenericInstantiationMixin):  # TODO: maybe a PackageBase class is needed to share members.
-	_packageReference: PackageReferenceSymbol
+	_packageReference:        PackageReferenceSymbol
 	_genericAssociationItems: List[GenericAssociationItem]
 
 	def __init__(
 		self,
-		identifier:          str,
-		genericPackage:      PackageReferenceSymbol,
-		contextItems:        Nullable[Iterable[ContextUnion]] =           None,
+		identifier:              str,
+		genericPackage:          PackageReferenceSymbol,
+		contextItems:            Nullable[Iterable[ContextUnion]] =           None,
 		genericAssociationItems: Nullable[Iterable[GenericAssociationItem]] = None,
-		documentation:       Nullable[str] =                              None,
-		parent:              Nullable[ModelEntity] =                      None
+		documentation:           Nullable[str] =                              None,
+		parent:                  Nullable[ModelEntity] =                      None
 	) -> None:
 		super().__init__(identifier, contextItems, documentation=documentation, parent=parent)
 		GenericEntityInstantiationMixin.__init__(self)
--- a/pyVHDLModel/__init__.py
+++ b/pyVHDLModel/__init__.py
@@ -256,9 +256,8 @@ class VHDLVersion(Enum):
 
 	def __hash__(self) -> int:
 		"""
-		Compute a hash value from the underlying version number, so ``VHDLVersion`` members can be used as
-		dict keys / set members again (overriding ``__eq__`` implicitly disables the default ``__hash__``).
-
+		Return the hash of the VHDL version using the underlying version number.
+		
 		.. note::
 
 		   ``Any`` compares equal to every other member (see :meth:`__eq__`), which no hash value can satisfy
@@ -266,7 +265,7 @@ class VHDLVersion(Enum):
 		   hashes by ``self.value``, which is internally consistent for all comparisons *except* those
 		   involving ``Any`` - avoid using ``Any`` as a dict key or set member.
 
-		:returns: Hash value of the underlying version number.
+		:returns: Hash value of the underlying VHDL version number.
 		"""
 		return hash(self.value)
 
--- a/pyVHDLModel/__init__.py
+++ b/pyVHDLModel/__init__.py
@@ -254,6 +254,22 @@ class VHDLVersion(Enum):
 		else:
 			raise TypeError("Second operand is not of type 'VHDLVersion'.")
 
+	def __hash__(self) -> int:
+		"""
+		Compute a hash value from the underlying version number, so ``VHDLVersion`` members can be used as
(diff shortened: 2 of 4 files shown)


pyVHDLModel/Instantiation.py
```diff
@@ -99,18 +99,18 @@ def __init__(
...
 
 		# TODO: extract to mixin
-		self._genericAssociations = []
-		if genericAssociations is not None:
-			for association in genericAssociations:
-				self._genericAssociations.append(association)
+		self._genericAssociationItems = []
+		if genericAssociationItems is not None:
+			for association in genericAssociationItems:
+				self._genericAssociationItems.append(association)
```
Paebbels, line 105: Don't we need to set `.Parent = self` for each assoziation?

pyVHDLModel/Instantiation.py
```diff
@@ -81,14 +81,14 @@ class FunctionInstantiation(Function, SubprogramInstantiationMixin):
...
-	_genericAssociations: List[GenericAssociationItem]
+	_genericAssociationItems: List[GenericAssociationItem]
 
 	def __init__(
 		self,
 		identifier:          str,
 		genericPackage:      PackageReferenceSymbol,
 		contextItems:        Nullable[Iterable[ContextUnion]] =           None,
-		genericAssociations: Nullable[Iterable[GenericAssociationItem]] = None,
+		genericAssociationItems: Nullable[Iterable[GenericAssociationItem]] = None,
```
Paebbels, line 91: Check all modified lines especially when touching parameters for alignments of types and default values if these exist.

pyVHDLModel/Instantiation.py
Paebbels: Already fixed via #128 (merged) - rebased this branch onto current dev to pick it up.

pyVHDLModel/Instantiation.py
Paebbels: Fixed - re-aligned PackageInstantiation's fields/params (the rename made genericAssociationItems the longest name/type, breaking the existing column alignment). Also went through the rest of this PR's new code with the same eye: aligned all 5 new multi-line constructors in Subprogram.py (Subprogram/Procedure/Function/ProcedureMethod/FunctionMethod) and ReturnStatement.__init__ in Sequential.py - none of those were aligned at all when I originally wrote them.

pyVHDLModel/__init__.py
```diff
@@ -254,6 +254,22 @@ def __eq__(self, other: Any) -> bool:
...
+		.. note::
+
+		   ``Any`` compares equal to every other member (see :meth:`__eq__`), which no hash value can satisfy
+		   simultaneously for all members without collapsing every member to the same hash. This implementation
+		   hashes by ``self.value``, which is internally consistent for all comparisons *except* those
+		   involving ``Any`` - avoid using ``Any`` as a dict key or set member.
+
+		:returns: Hash value of the underlying version number.
+		"""
+		return hash(self.value)
```
Paebbels, line 270:
```suggestion
	def __hash__(self) -> int:
		"""
		Return the hash of the VHDL version using the underlying version number.
		
		.. note::

		   ``Any`` compares equal to every other member (see :meth:`__eq__`), which no hash value can satisfy
		   simultaneously for all members without collapsing every member to the same hash. This implementation
		   hashes by ``self.value``, which is internally consistent for all comparisons *except* those
		   involving ``Any`` - avoid using ``Any`` as a dict key or set member.

		:returns: Hash value of the underlying VHDL version number.
		"""
		return hash(self.value)
```

PR #134, opened 2026-07-21
Add VHDL-2019 mode view classes and split Port/Parameter signal interface items

Paebbels:
## Summary

VHDL-2019 mode views. Design was discussed and agreed on in chat first (including cross-checking the actual grammar rather than relying on memory), then implemented.

### New classes

- `ModeViewSymbol` (`Symbol.py`) - reference to a mode view declaration, mirroring `PackageReferenceSymbol` exactly. Uses `PossibleReference.View`, which - along with `.ViewAttribute` - was already present in the enum.
- `ModeViewDeclaration`, `ModeViewElement`, `SimpleModeViewElement`, `CompositeModeViewElement` (`Interface.py`):
  - `ModeViewElement` uses `MultipleNamedEntityMixin` - mode view elements support comma-separated identifiers sharing one mode (`a, b : out;`), same pattern as `Constant`/`Signal`/`Variable`.
  - `CompositeModeViewElement` merges what GHDL's IIR splits into `Array_Mode_View_Element`/`Record_Mode_View_Element` - verified both are structurally identical (just a reference to another named view); the distinction requires resolving the target field's type, which needs semantic analysis this project doesn't perform.
  - `ModeViewDeclaration.Subtype` is a plain `SubtypeSymbol`, so `of MyArrayType(0 to 7)` (constrained array-of-views) is already supported for free via the existing `ConstrainedArraySubtypeSymbol` - no new work needed there.

### Port/Parameter interface item split

Split `PortSignalInterfaceItem` and `ParameterSignalInterfaceItem` into abstract bases with two concrete forms each (`*SimpleSignalInterfaceItem` / `*ViewSignalInterfaceItem`), since mode views apply to signal-class ports and signal-class subprogram parameters, but not to generics.

Cross-checked against the actual VHDL grammar (not just memory/LRM prose) before finalizing this: `interface_object_declaration` is one production shared across generic/port/parameter lists - the generics-can't-be-signals restriction is semantic (LRM prose), not encoded in the grammar itself. This explains why GHDL's parser (grammar-level only, no full semantic pass) accepts view-typed generics/parameters syntactically without them being legal VHDL - confirmed empirically against real GHDL, then reconciled against the grammar and confirmed as correct by the model owner directly. `GenericSignalInterfaceItem` does not exist and is out of scope here.

### One small supporting change

`Obj.__init__` (`Object.py`) now tolerates `subtype=None` (`Nullable[Symbol]`, additive, not breaking). A view-typed port has no separate subtype indication of its own at the parse-only level this project operates at - verified against real GHDL: `Get_Subtype_Indication` on `Interface_View_Declaration` is `Null_Iir` before semantic analysis resolves the referenced view. The type is only implied by the mode view.

## Testing

Added `tests/unit/Interface.py`. Full suite: `82 passed` (was 73), no regressions.

## Companion change

pyGHDL.dom will need a matching PR to actually translate `Mode_View_Declaration`/`Interface_View_Declaration` using these classes, plus `Design`'s `--std=19` support (already merged) to reach this code path at all.

### DIFF
--- a/pyVHDLModel/Interface.py
+++ b/pyVHDLModel/Interface.py
@@ -278,6 +278,7 @@ class PortSignalInterfaceItem(Signal, InterfaceItemMixin):
 @export
 class PortSimpleSignalInterfaceItem(PortSignalInterfaceItem, InterfaceItemWithModeMixin):
 	"""
+
 	.. admonition:: Example
 
 	   .. code-block:: VHDL
@@ -301,6 +302,7 @@ class PortSimpleSignalInterfaceItem(PortSignalInterfaceItem, InterfaceItemWithMo
 @export
 class PortViewSignalInterfaceItem(PortSignalInterfaceItem):
 	"""
+
 	.. admonition:: Example
 
 	   .. code-block:: VHDL
@@ -309,29 +311,26 @@ class PortViewSignalInterfaceItem(PortSignalInterfaceItem):
 
 	.. note::
 
-	   ``Subtype`` will usually be ``None`` here: at the parse-only level this project operates at, the
-	   element's type is only implied by the referenced mode view's own ``of`` type - there is no separate
-	   subtype indication to read directly off the interface declaration itself.
+	   VHDL's grammar treats a mode view indication as occupying the same structural position as an
+	   ordinary subtype indication (``mode_indication ::= simple_mode_indication | mode_view_indication``).
+	   Accordingly, the mode view reference *is* this object's :attr:`Subtype` (a :class:`Symbol` is a
+	   :class:`Symbol`, whether it names a subtype or a mode view) - :attr:`ModeViewIndication` is just a
+	   more specific, aliased name for the same value, not a separate field. An object's subtype can never
+	   be ``None`` - there is no VHDL syntax that omits it.
 	"""
-
-	_modeViewIndication: ModeViewSymbol
 
 	def __init__(
 		self,
 		identifiers: Iterable[str],
 		modeViewIndication: ModeViewSymbol,
-		subtype: Nullable[Symbol] = None,
 		documentation: Nullable[str] = None,
 		parent: Nullable[ModelEntity] = None
 	) -> None:
-		super().__init__(identifiers, subtype, None, documentation, parent)
-
-		self._modeViewIndication = modeViewIndication
-		modeViewIndication.Parent = self
+		super().__init__(identifiers, modeViewIndication, None, documentation, parent)
 
 	@readonly
 	def ModeViewIndication(self) -> ModeViewSymbol:
-		return self._modeViewIndication
+		return self._subtype
 
 
 @export
@@ -378,6 +377,7 @@ class ParameterSignalInterfaceItem(Signal, ParameterInterfaceItemMixin):
 @export
 class ParameterSimpleSignalInterfaceItem(ParameterSignalInterfaceItem, InterfaceItemWithModeMixin):
 	"""
+
 	.. admonition:: Example
 
 	   .. code-block:: VHDL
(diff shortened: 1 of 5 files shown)


pyVHDLModel/Interface.py
```diff
@@ -39,14 +39,135 @@
...
+class CompositeModeViewElement(ModeViewElement):
+	"""
+	A mode view element that refers to another (named) mode view for an array or record sub-element.
+
+	GHDL's IIR distinguishes an array- from a record-typed target element (``Array_Mode_View_Element`` /
+	``Record_Mode_View_Element``), but that distinction can only be determined once the target element's
+	type is resolved, which requires semantic analysis. Both are merged into this one class here, since they
+	are otherwise structurally identical (both just carry a reference to another mode view).
+
+	.. admonition:: Example
```
```suggestion

	.. admonition:: Example
```

pyVHDLModel/Object.py
```diff
@@ -59,21 +59,22 @@ class Obj(ModelEntity, MultipleNamedEntityMixin, DocumentedEntityMixin):
 	:data:`__objectVertex`.
 	"""
 
-	_subtype:      Symbol
+	_subtype:      Nullable[Symbol]
```
Paebbels: The subtype for an object (signal, variable, constant, file) cannot be None. It must always exist. There is no syntax allowed where a subtype is omitted when creating an object.

Check again why your None error happend and find out the root cause. This is an unacceptable work around.

pyVHDLModel/Interface.py
```diff
@@ -146,7 +267,24 @@ def __init__(self, identifier: str, documentation: Nullable[str] = None, parent:
 
 
 @export
-class PortSignalInterfaceItem(Signal, PortInterfaceItemMixin):
+class PortSignalInterfaceItem(Signal, InterfaceItemMixin):
```
Paebbels, line 264: PortSignalInterfaceItem became an abstract base class. I think it was used in many places in the design, right? Shouldn't you modify more location to now use the new PortSimpleSignalInterfaceItem or is this change limited to the pyGHDL.dom side and not visible in pyVHDLModel?

pyVHDLModel/Interface.py
Paebbels: Fixed - and checked the whole file for the same issue (missed 3 more of the same pattern, all fixed now).

pyVHDLModel/Object.py
Paebbels: You're right, sorry - reverted this entirely. Root cause: Obj._subtype is typed as the general Symbol base class, not SubtypeSymbol specifically, and ModeViewSymbol is also a Symbol. VHDL's own grammar treats a mode view indication as occupying the same structural position as an ordinary subtype indication (mode_indication ::= simple_mode_indication | mode_view_indication). So PortViewSignalInterfaceItem/ParameterViewSignalInterfaceItem now pass the ModeViewSymbol itself as the real subtype value - Subtype is never None again, and ModeViewIndication is just an aliased, more specific name for the same value, not a separate field. Verified end-to-end through pyGHDL.dom against real GHDL output too, not just unit tests.

pyVHDLModel/Interface.py
Paebbels: Checked the whole repo - no other construction sites in pyVHDLModel itself (only the class definitions). But doc/LanguageModel/InterfaceItems.rst was indeed stale - it still showed PortSignalInterfaceItem/ParameterSignalInterfaceItem as concrete, matching the pre-split shape. Updated the inheritance diagram and added proper sections for the four new concrete classes. Left the file's other pre-existing staleness alone (e.g. the SyntaxModel module path references) - that predates this change.

pyVHDLModel/Interface.py
```diff
@@ -39,14 +39,135 @@
...
+@export
+class CompositeModeViewElement(ModeViewElement):
+	"""
+	A mode view element that refers to another (named) mode view for an array or record sub-element.
+
+	GHDL's IIR distinguishes an array- from a record-typed target element (``Array_Mode_View_Element`` /
+	``Record_Mode_View_Element``), but that distinction can only be determined once the target element's
+	type is resolved, which requires semantic analysis. Both are merged into this one class here, since they
+	are otherwise structurally identical (both just carry a reference to another mode view).
+
```
```suggestion

```

PR #135, opened 2026-07-22
Subprogram instantiation (VHDL-2008 generic subprograms) and Alias

Paebbels:
## Summary

Two independent fixes grouped together, as requested: subprogram instantiation, then alias.

### 1. Subprogram instantiation (VHDL-2008 generic subprograms)

`FunctionInstantiation`/`ProcedureInstantiation` were bare `pass` stub classes, and
`SubprogramInstantiationMixin.__init__` took no parameters at all - `_subprogramReference` was
always `None`, with no way to populate it (flagged in the code itself:
`# FIXME: is this a subprogram symbol?`).

- New `SubprogramReferenceSymbol` (`Symbol.py`), mirroring `PackageReferenceSymbol`, using the
  existing `PossibleReference.SubProgram` flag. Fixes the exact mismatch flagged in that FIXME -
  `_subprogramReference` was typed as the resolved `Subprogram` entity directly rather than a
  `Symbol` wrapper.
- `SubprogramInstantiationMixin` now properly accepts and stores `subprogramReference` and
  `genericAssociationItems` (mirroring `PackageInstantiation`'s pattern).
- `ProcedureInstantiation` gets a real constructor.
- `FunctionInstantiation` gets a real constructor too, but deliberately does **not** call
  `Function.__init__` (which requires a mandatory, non-`None` `returnType`) - it calls
  `Subprogram.__init__` directly and sets `_returnType = None`.

  This isn't a repeat of the `Obj.Subtype` mistake from #134: verified against real GHDL that
  `Get_Return_Type` on a `Function_Instantiation_Declaration` is `Null_Iir`, and - unlike a plain
  function - the LRM grammar never lets a subprogram instantiation write its own return type at
  all. It can only ever be known by resolving the referenced uninstantiated subprogram, which needs
  semantic analysis. `None` here reflects a not-yet-resolved reference, the same status as any
  other unresolved `Symbol.Reference`, not an omission the source syntax would never allow (which is
  what made the `Obj.Subtype` case wrong).

### 2. Alias

`Alias` had no field at all for what's being aliased - only its own identifier and documentation.
`alias b is s;` lost the fact that `b` aliases `s` entirely.

- `Alias.Name` (a plain `Name`, not a `*ReferenceSymbol` - an alias can refer to almost anything
  nameable: objects, types, subprograms, literals, so none of the narrower `ReferenceSymbol` classes
  fit).
- `Alias.Subtype` (`Nullable[SubtypeSymbol]` - genuinely optional per the grammar:
  `alias_designator [ : subtype_indication ] IS name [ signature ]`, unlike `Obj.Subtype` which is
  never optional).

## Testing

Added `tests/unit/Declaration.py` (Alias) and extended `tests/unit/Instantiate.py` (subprogram
instantiation). Full suite: `86 passed` (was 82), no regressions.

## Companion change

pyGHDL.dom PR translating both, using these classes.

### DIFF
--- a/pyVHDLModel/Declaration.py
+++ b/pyVHDLModel/Declaration.py
@@ -182,10 +182,15 @@ class Alias(ModelEntity, NamedEntityMixin, DocumentedEntityMixin):
 	"""
 	Represents an alias declaration.
 
-	An alias can refer to almost anything nameable (an object, a type, a subprogram, a literal, ...), so
-	:attr:`Name` is a plain :class:`~pyVHDLModel.Name.Name` rather than one of the narrower
-	``*ReferenceSymbol`` classes (those are for cases where the kind of thing being referenced is known
-	up front, e.g. a package or a mode view).
+	:attr:`Name` is a :class:`~pyVHDLModel.Symbol.Symbol` - like every other cross-reference in this
+	model - rather than a bare :class:`~pyVHDLModel.Name.Name`, so it participates in the usual
+	resolve-later mechanism (:attr:`~pyVHDLModel.Symbol.Symbol.Reference` /
+	:attr:`~pyVHDLModel.Symbol.Symbol.IsResolved`). Unlike ``PackageReferenceSymbol`` and similar,
+	there is no single fixed :class:`~pyVHDLModel.Symbol.PossibleReference` value that always fits: an
+	alias without a subtype indication can refer to almost anything nameable (an object, a type, a
+	subprogram, a literal, ...), while an alias *with* a subtype indication can - per the LRM - only
+	ever refer to an object (a constant, variable, signal, or file); the ``possibleReferences`` passed
+	to the ``Symbol`` should reflect whichever case applies.
 
 	.. admonition:: Example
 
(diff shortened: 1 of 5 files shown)


pyVHDLModel/Declaration.py
```diff
@@ -179,12 +179,53 @@ def Expression(self) -> ExpressionUnion:
...
+	      alias a : bit_vector(3 downto 0) is s(3 downto 0);
+	      --        ^^^^^^^^^^^^^^^^^^^^^^     ^^^^^^^^^^^^
+	      --        Subtype (optional)         Name
+
+	      alias b is s;
+	      --          ^
+	      --          Name
+	"""
+
+	_name:    Name
```
Paebbels: Isn't this name a Symbol? In case of an existing subtype, it can only be a object (signal, variable, constant, file) that's referenced, right?

Paebbels: Good catch, fixed - Alias.Name is now a Symbol. Confirmed your point too: with an explicit subtype indication, the LRM restricts an alias to referencing an object specifically, so PossibleReference.Object (which already existed) is used in that case. Without a subtype, the target can be almost anything nameable, so a broader PackageMember | EnumLiteral is used there instead.

pyGHDL.dom companion change adds a generic Symbol DOM wrapper (no existing class fit, since none of the specific *ReferenceSymbol classes are right for something this open-ended) and updates Alias.parse() to construct it with the appropriate PossibleReference depending on whether a subtype indication is present.

Verified against real GHDL analysis - all four alias variants still translate correctly. Full suite: 86 passed (pyVHDLModel), 41 passed/5 skipped (pyGHDL.dom).

PR #136, opened 2026-07-22
Fix ConstrainedScalarSubtypeSymbol: range constraint had nowhere to go

Paebbels:
## Summary

`signal s : integer range 0 to 15;` parses fine, but `ConstrainedScalarSubtypeSymbol` was a bare
stub (`class ConstrainedScalarSubtypeSymbol(SubtypeSymbol): pass`) - no fields, no constructor
parameter for a range at all. pyGHDL.dom's `ConstrainedScalarSubtypeSymbol.__init__` took an
`rng: Range` parameter but never forwarded it
(`super().__init__(subtypeName)  # , rng)  # XXX: hacked`), and its `.parse()` classmethod was a
bare `pass`.

## Changes

Added `ScalarConstraint` (mirroring `ArrayConstraint`/`RecordConstraint`'s established pattern for
consistency, even though - like those two - it's not reused elsewhere) and gave
`ConstrainedScalarSubtypeSymbol` a real constructor, storing the range via
`ScalarConstraint.Constraint`.

`Constraint` is `Nullable` - not because the source syntax ever omits a range constraint for a
constrained scalar subtype (it never does), but because pyGHDL.dom doesn't yet extract a range from
an `AttributeName`-based constraint (e.g. `subtype s is t'range;`) - a pre-existing, separately-
tracked TODO in that function, not something addressed here.

## Testing

Added tests to `tests/unit/Instantiate.py`. Full suite: `88 passed` (was 86), no regressions.

## Companion change

pyGHDL.dom PR forwarding the range constraint properly using this fix.

### DIFF
--- a/pyVHDLModel/Symbol.py
+++ b/pyVHDLModel/Symbol.py
@@ -507,12 +507,12 @@ class SimpleSubtypeSymbol(SubtypeSymbol):
 
 
 @export
-class Constraint:
+class Constraint(metaclass=ExtendedType, mixin=True):
 	pass
 
 
 @export
-class ScalarConstraint(Constraint):
+class ScalarConstraint(Constraint, mixin=True):
 	_constraint: Nullable[Range]
 
 	def __init__(self, constraint: Nullable[Range]) -> None:
@@ -548,7 +548,7 @@ class ConstrainedScalarSubtypeSymbol(SubtypeSymbol, ScalarConstraint):
 
 
 @export
-class ArrayConstraint(Constraint):
+class ArrayConstraint(Constraint, mixin=True):
 	_constraints: List[Range]
 
 	def __init__(self, constraints: Iterable[Range]) -> None:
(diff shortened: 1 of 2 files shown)


pyVHDLModel/Symbol.py
```diff
@@ -507,13 +507,44 @@ class SimpleSubtypeSymbol(SubtypeSymbol):
 
 
 @export
-class ConstrainedScalarSubtypeSymbol(SubtypeSymbol):
+class Constraint:
```
Paebbels: Don't we need a base-class like ModelEntity? Look at the other very basics like a design unit or an object (later derived to a signal or variable).

Alternatively, if this is a mixin, see other mixins. These should use all the same common `metaclass=ExtendedType` and a metaclass parameter `mixin=True`.

Paebbels: Good catch - fixed all four (Constraint, ScalarConstraint, ArrayConstraint, RecordConstraint) to use metaclass=ExtendedType, mixin=True, matching the convention every other mixin in this codebase follows. This was silently working before only because Python resolves the metaclass automatically when one base (SubtypeSymbol, via ExtendedType) is more derived than the plain default type metaclass on the other - not because the declaration was actually correct.

PR #138, opened 2026-07-22
Add conditional/selected/force/release assignment statements

Paebbels:
## Summary

Design discussed and agreed on in chat first, then implemented. Covers the remaining gap in both
the concurrent and sequential statement dispatchers - everything else in both was already complete
(confirmed by a full audit of both dispatchers before starting).

## New shared building blocks (`Common.py`)

Used by both concurrent and sequential forms, since GHDL represents the sequential and concurrent
conditional-signal-assignment/selected-signal-assignment forms with the exact same underlying
structure:

- `ConditionalWaveform` / `ConditionalExpression`: one branch of a conditional assignment
  (waveform-or-expression + optional condition; `None` condition = the final "else" branch).
- `SelectedWaveform` / `OthersSelectedWaveform`, `SelectedExpression` / `OthersSelectedExpression`:
  one alternative of a selected assignment, mirroring the `GenerateCase`/`OthersGenerateCase`
  pattern (choices + associated content) rather than modeling "others" as a special choice value,
  since a selected-assignment "others" alternative - like a generate-case "others" - can't be
  combined with other choices via `|` and is better represented as a distinct alternative kind.

## Fixed two existing stubs (`Concurrent.py`)

Both previously accepted an `expression` parameter not actually suited to what they need to store:
- `ConcurrentConditionalSignalAssignment`: now holds `ConditionalWaveforms`.
- `ConcurrentSelectedSignalAssignment`: now holds the select expression + `SelectedWaveforms`.

## Six new classes (`Sequential.py`)

- `SequentialConditionalVariableAssignment` / `SequentialConditionalSignalAssignment` (VHDL-2008)
- `SequentialSelectedVariableAssignment` / `SequentialSelectedSignalAssignment`
- `SignalForceAssignment` / `SignalReleaseAssignment` (VHDL-2008)

**Note**: `SequentialConditionalVariableAssignment` and `SequentialSelectedVariableAssignment`
inherit `AssignmentMixin` (`Target` only), not `VariableAssignmentMixin` - the latter also promises
a single `Expression` property backed by a single stored value, which doesn't fit either class
(they hold a list of conditional expressions, or a select expression plus a list of selected
expressions). Caught this while implementing, before it became a live `AttributeError` on first
access.

`SequentialVariableAssignment` (simple, non-conditional/selected) was already fully implemented -
not touched here, only the two more elaborate sequential variable-assignment forms are new.

## Testing

Added `tests/unit/Assignment.py`, covering all 9 statement classes plus the building blocks. Full
suite: `117 passed` (was 101), no regressions.

## Companion change

pyGHDL.dom PR translating all 9 statement kinds using these classes.

### DIFF
--- a/pyVHDLModel/Common.py
+++ b/pyVHDLModel/Common.py
@@ -357,3 +357,44 @@ class OthersSelectedExpression(BaseCase, ExpressionMixin):
 	def __init__(self, expression: ExpressionUnion, parent: Nullable[ModelEntity] = None) -> None:
 		super().__init__(parent)
 		ExpressionMixin.__init__(self, expression)
+
+
+@export
+class SelectedWaveformsMixin(metaclass=ExtendedType, mixin=True):
+	"""
+	A mixin-class for all statements holding a list of :class:`SelectedWaveform`/
+	:class:`OthersSelectedWaveform` (both the concurrent and sequential forms of a selected signal
+	assignment).
+	"""
+
+	_selectedWaveforms: List[Union[SelectedWaveform, OthersSelectedWaveform]]
+
+	def __init__(self, selectedWaveforms: Iterable[Union[SelectedWaveform, OthersSelectedWaveform]]) -> None:
+		self._selectedWaveforms = []
+		for selectedWaveform in selectedWaveforms:
+			self._selectedWaveforms.append(selectedWaveform)
+			selectedWaveform.Parent = self
+
+	@readonly
+	def SelectedWaveforms(self) -> List[Union[SelectedWaveform, OthersSelectedWaveform]]:
+		return self._selectedWaveforms
+
+
+@export
+class SelectedExpressionsMixin(metaclass=ExtendedType, mixin=True):
+	"""
+	A mixin-class for all statements holding a list of :class:`SelectedExpression`/
+	:class:`OthersSelectedExpression`.
+	"""
+
+	_selectedExpressions: List[Union[SelectedExpression, OthersSelectedExpression]]
+
+	def __init__(self, selectedExpressions: Iterable[Union[SelectedExpression, OthersSelectedExpression]]) -> None:
+		self._selectedExpressions = []
+		for selectedExpression in selectedExpressions:
+			self._selectedExpressions.append(selectedExpression)
+			selectedExpression.Parent = self
+
+	@readonly
+	def SelectedExpressions(self) -> List[Union[SelectedExpression, OthersSelectedExpression]]:
+		return self._selectedExpressions
--- a/pyVHDLModel/Common.py
+++ b/pyVHDLModel/Common.py
@@ -41,7 +41,7 @@ from pyTooling.MetaClasses   import ExtendedType
 
 from pyVHDLModel.Base        import ModelEntity, LabeledEntityMixin, BaseCase, BaseChoice, WaveformElement
 from pyVHDLModel.Expression  import BaseExpression, QualifiedExpression, FunctionCall, TypeConversion, Literal
-from pyVHDLModel.Symbol      import Symbol
+from pyVHDLModel.Symbol      import Symbol, SignalSymbol, VariableSymbol
 from pyVHDLModel.Association import ParameterAssociationItem
 
 
@@ -144,6 +144,10 @@ class AssignmentMixin(metaclass=ExtendedType, mixin=True):
 class SignalAssignmentMixin(AssignmentMixin, mixin=True):
 	"""A mixin-class for all signal assignment statements."""
 
+	@property
+	def Target(self) -> SignalSymbol:
+		return self._target
+
 
 @export
 class VariableAssignmentMixin(AssignmentMixin, mixin=True):
@@ -152,11 +156,15 @@ class VariableAssignmentMixin(AssignmentMixin, mixin=True):
 	# FIXME: move to sequential?
 	_expression: ExpressionUnion
 
-	def __init__(self, target: Symbol, expression: ExpressionUnion) -> None:
+	def __init__(self, target: VariableSymbol, expression: ExpressionUnion) -> None:
 		super().__init__(target)
 
 		self._expression = expression
 		expression.Parent = self
+
+	@property
+	def Target(self) -> VariableSymbol:
+		return self._target
 
 	@property
 	def Expression(self) -> ExpressionUnion:
@@ -245,6 +253,26 @@ class ConditionalExpression(ModelEntity):
 	@readonly
 	def Condition(self) -> Nullable[ExpressionUnion]:
 		return self._condition
+
+
+@export
+class ConditionalWaveformsMixin(metaclass=ExtendedType, mixin=True):
+	"""
+	A mixin-class for all statements holding a list of :class:`ConditionalWaveform` (both the
+	concurrent and sequential forms of a conditional signal assignment).
+	"""
+
+	_conditionalWaveforms: List[ConditionalWaveform]
+
+	def __init__(self, conditionalWaveforms: Iterable[ConditionalWaveform]) -> None:
+		self._conditionalWaveforms = []
+		for conditionalWaveform in conditionalWaveforms:
+			self._conditionalWaveforms.append(conditionalWaveform)
+			conditionalWaveform.Parent = self
+
+	@readonly
+	def ConditionalWaveforms(self) -> List[ConditionalWaveform]:
+		return self._conditionalWaveforms
 
 
 @export
--- a/pyVHDLModel/Common.py
+++ b/pyVHDLModel/Common.py
@@ -39,7 +39,7 @@ from typing                  import List, Iterable, Union, Optional as Nullable
 from pyTooling.Decorators    import export, readonly
 from pyTooling.MetaClasses   import ExtendedType
 
-from pyVHDLModel.Base        import ModelEntity, LabeledEntityMixin, BaseCase, BaseChoice, WaveformElement, ConditionalMixin
+from pyVHDLModel.Base        import ModelEntity, LabeledEntityMixin, BaseCase, BaseChoice, WaveformElement, ConditionalMixin, ChoicesMixin
 from pyVHDLModel.Expression  import BaseExpression, QualifiedExpression, FunctionCall, TypeConversion, Literal
 from pyVHDLModel.Symbol      import Symbol, SignalSymbol, VariableSymbol
 from pyVHDLModel.Association import ParameterAssociationItem
@@ -274,7 +274,7 @@ class ConditionalWaveformsMixin(metaclass=ExtendedType, mixin=True):
 
 
 @export
-class SelectedWaveform(BaseCase, WaveformMixin):
+class SelectedWaveform(BaseCase, WaveformMixin, ChoicesMixin):
 	"""
 	One alternative of a selected (waveform) signal assignment.
 
@@ -286,8 +286,6 @@ class SelectedWaveform(BaseCase, WaveformMixin):
 	      --                    ^^^^^^^^      <- this alternative: Choices=[0], Waveform=['1']
 	"""
 
-	_choices: List[BaseChoice]
-
 	def __init__(
 		self,
 		choices: Iterable[BaseChoice],
@@ -296,15 +294,7 @@ class SelectedWaveform(BaseCase, WaveformMixin):
 	) -> None:
 		super().__init__(parent)
 		WaveformMixin.__init__(self, waveform)
-
-		self._choices = []
-		for choice in choices:
-			self._choices.append(choice)
-			choice.Parent = self
-
-	@readonly
-	def Choices(self) -> List[BaseChoice]:
-		return self._choices
+		ChoicesMixin.__init__(self, choices)
 
 
 @export
@@ -317,7 +307,7 @@ class OthersSelectedWaveform(BaseCase, WaveformMixin):
 
 
 @export
-class SelectedExpression(BaseCase, ExpressionMixin):
+class SelectedExpression(BaseCase, ExpressionMixin, ChoicesMixin):
 	"""
 	One alternative of a selected (variable) assignment.
 
@@ -329,8 +319,6 @@ class SelectedExpression(BaseCase, ExpressionMixin):
 	      --                   ^^^^^^^^      <- this alternative: Choices=[0], Expression='1'
 	"""
 
-	_choices: List[BaseChoice]
-
 	def __init__(
 		self,
 		choices: Iterable[BaseChoice],
@@ -339,15 +327,7 @@ class SelectedExpression(BaseCase, ExpressionMixin):
 	) -> None:
 		super().__init__(parent)
 		ExpressionMixin.__init__(self, expression)
-
-		self._choices = []
-		for choice in choices:
-			self._choices.append(choice)
-			choice.Parent = self
-
-	@readonly
-	def Choices(self) -> List[BaseChoice]:
-		return self._choices
+		ChoicesMixin.__init__(self, choices)
 
 
 @export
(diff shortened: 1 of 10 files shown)


pyVHDLModel/Concurrent.py
```diff
@@ -929,15 +930,74 @@ def Waveform(self) -> List[WaveformElement]:
...
+	      with sel select s <= '1' when 0, '0' when others;
+	"""
+
+	_expression:        ExpressionUnion
+	_selectedWaveforms: List[SelectedWaveform]
+
+	def __init__(
+		self,
+		label: str,
+		target: Name,
```
Paebbels: Don't we need a symbol here? I think it also must be a SignalSymbol (maybe needs to be created similar to a VariableSymbol in variable assignments). Check other targets too.

pyVHDLModel/Sequential.py
```diff
@@ -125,6 +128,197 @@ def __init__(self, target: Symbol, expression: ExpressionUnion, label: Nullable[
...
+	   .. code-block:: VHDL
+
+	      s <= '1' when cond1 else '0' when cond2 else 'Z';
+	"""
+
+	_conditionalWaveforms: List[ConditionalWaveform]
+
+	def __init__(
+		self,
+		target: Symbol,
```
Paebbels: Here it is a symbol, but it could be more specific, because it's an assignment to a signal.

pyVHDLModel/Concurrent.py
```diff
@@ -929,15 +930,74 @@ def Waveform(self) -> List[WaveformElement]:
...
+		conditionalWaveforms: Iterable[ConditionalWaveform],
+		parent: Nullable[ModelEntity] = None
+	) -> None:
 		super().__init__(label, target, parent)
 
+		self._conditionalWaveforms = []
+		for conditionalWaveform in conditionalWaveforms:
+			self._conditionalWaveforms.append(conditionalWaveform)
+			conditionalWaveform.Parent = self
```
Paebbels: Can we extract that part into a mixin (make sure to use ExtendedType and mixin=True)?

pyVHDLModel/Sequential.py
```diff
@@ -125,6 +128,197 @@ def __init__(self, target: Symbol, expression: ExpressionUnion, label: Nullable[
...
+		SignalAssignmentMixin.__init__(self, target)
+
+		self._conditionalWaveforms = []
+		for conditionalWaveform in conditionalWaveforms:
+			self._conditionalWaveforms.append(conditionalWaveform)
+			conditionalWaveform.Parent = self
+
+	@readonly
+	def ConditionalWaveforms(self) -> List[ConditionalWaveform]:
+		return self._conditionalWaveforms
```
Paebbels: This could also go into the mixin.

tests/unit/Assignment.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+if __name__ == "__main__":  # pragma: no cover
+	print("ERROR: you called a testcase declaration file as an executable module.")
+	print("Use: 'python -m unitest <testcase module>'")
+	exit(1)
+
+
+def _target() -> Symbol:
+	return Symbol(SimpleName("s"), PossibleReference.Signal)
```
Paebbels: By introducing SignalSymbol, this shouldn't be needed anymore.

tests/unit/Assignment.py
```diff
@@ -0,0 +1,193 @@
...
+#  |_|    |___/                                                                                                        #
+# ==================================================================================================================== #
+# Authors:                                                                                                             #
+#   Patrick Lehmann                                                                                                    #
+#                                                                                                                      #
+# License:                                                                                                             #
+# ==================================================================================================================== #
+# Copyright 2017-2026 Patrick Lehmann - Boetzingen, Germany                                                            #
+# Copyright 2016-2017 Patrick Lehmann - Dresden, Germany                                                               #
```
Paebbels: All new files added since v0.37.0 should use:
# Copyright 2026-2026 Patrick Lehmann - Boetzingen, Germany
Please check git history for all added files since last release (tag v0.37.0) on main branch.

pyVHDLModel/Common.py
```diff
@@ -144,6 +144,10 @@ def Target(self) -> Symbol:
 class SignalAssignmentMixin(AssignmentMixin, mixin=True):
 	"""A mixin-class for all signal assignment statements."""
 
+	@property
```
Paebbels: This is a readonly property, thus use `@readonly`.

pyVHDLModel/Common.py
```diff
@@ -152,12 +156,240 @@ class VariableAssignmentMixin(AssignmentMixin, mixin=True):
...
 	_expression: ExpressionUnion
 
-	def __init__(self, target: Symbol, expression: ExpressionUnion) -> None:
+	def __init__(self, target: VariableSymbol, expression: ExpressionUnion) -> None:
 		super().__init__(target)
 
 		self._expression = expression
 		expression.Parent = self
 
 	@property
```
Paebbels: `@readonly`

pyVHDLModel/Common.py
```diff
@@ -152,12 +156,240 @@ class VariableAssignmentMixin(AssignmentMixin, mixin=True):
...
+	def Target(self) -> VariableSymbol:
+		return self._target
+
+	@property
```
Paebbels: `@readonly`

pyVHDLModel/Common.py
```diff
@@ -152,12 +156,240 @@ class VariableAssignmentMixin(AssignmentMixin, mixin=True):
...
+		super().__init__(parent)
+
+		self._waveform = []
+		for waveformElement in waveform:
+			self._waveform.append(waveformElement)
+			waveformElement.Parent = self
+
+		self._condition = condition
+		if condition is not None:
+			condition.Parent = self
```
Paebbels: for this and the matching Condition readonly property, don't we have a ConditionalMixin or so? I guess it was used with generate and if statements.

pyVHDLModel/Common.py
```diff
@@ -152,12 +156,240 @@ class VariableAssignmentMixin(AssignmentMixin, mixin=True):
...
+		waveform: Iterable[WaveformElement],
+		condition: Nullable[ExpressionUnion] = None,
+		parent: Nullable[ModelEntity] = None
+	) -> None:
+		super().__init__(parent)
+
+		self._waveform = []
+		for waveformElement in waveform:
+			self._waveform.append(waveformElement)
+			waveformElement.Parent = self
```
Paebbels: this can be extracted too including the matching Waveform readonly property.

pyVHDLModel/Common.py
```diff
@@ -152,12 +156,240 @@ class VariableAssignmentMixin(AssignmentMixin, mixin=True):
...
+
+	@readonly
+	def Choices(self) -> List[BaseChoice]:
+		return self._choices
+
+	@readonly
+	def Waveform(self) -> List[WaveformElement]:
+		return self._waveform
```
Paebbels: See comment above for extracting into a mixin.

pyVHDLModel/Common.py
```diff
@@ -152,12 +156,240 @@ class VariableAssignmentMixin(AssignmentMixin, mixin=True):
...
+		super().__init__(parent)
+
+		self._waveform = []
+		for waveformElement in waveform:
+			self._waveform.append(waveformElement)
+			waveformElement.Parent = self
+
+	@readonly
+	def Waveform(self) -> List[WaveformElement]:
+		return self._waveform
```
Paebbels: again this could be extracted as mentioned above.

pyVHDLModel/Concurrent.py
Paebbels: Added SignalSymbol (mirroring ConfigurationSymbol, using PossibleReference.Signal) and retyped every target across Concurrent.py and Sequential.py (9 constructors) to the correct one - SignalSymbol for signal assignments, VariableSymbol for variable assignments. Also narrowed .Target's return type on SignalAssignmentMixin/VariableAssignmentMixin. Companion pyGHDL.dom PR updated to actually construct these instead of a bare Name.

pyVHDLModel/Sequential.py
Paebbels: Fixed - see the reply on the Concurrent.py comment, same change applies here (VariableSymbol for the variable-assignment target).

pyVHDLModel/Concurrent.py
Paebbels: Extracted ConditionalWaveformsMixin into Common.py (metaclass=ExtendedType, mixin=True) - both ConcurrentConditionalSignalAssignment and SequentialConditionalSignalAssignment now use it instead of duplicating the storage/iteration logic.

pyVHDLModel/Sequential.py
Paebbels: Done - now uses the same ConditionalWaveformsMixin from Common.py.

tests/unit/Assignment.py
Paebbels: Right - removed the generic Symbol usage, tests now use _signalTarget()/_variableTarget() helpers built on the new SignalSymbol/VariableSymbol classes.

tests/unit/Assignment.py
Paebbels: Fixed - checked git history for all files added since v0.37.0 (dev has advanced past it even though main hasn't yet) and corrected the header on all 6: pyVHDLModel/Configuration.py, tests/unit/Assignment.py, tests/unit/Configuration.py, tests/unit/Declaration.py, tests/unit/Interface.py (tests/unit/Hierarchy.py already had it right).

pyVHDLModel/Common.py
Paebbels: Fixed. Also caught and fixed the same issue on AssignmentMixin.Target and VariableAssignmentMixin.Target/.Expression while I was at it.

pyVHDLModel/Common.py
Paebbels: Fixed.

pyVHDLModel/Common.py
Paebbels: Fixed.

pyVHDLModel/Common.py
Paebbels: Good catch - now reuses the existing ConditionalMixin (Base.py, already used for if/elsif branches) for the Condition part of both ConditionalWaveform and ConditionalExpression, instead of duplicating it.

pyVHDLModel/Common.py
Paebbels: Extracted into a new WaveformMixin (Common.py). Also used it to resolve the two pre-existing '# TODO: extract to mixin' comments on ConcurrentSimpleSignalAssignment/SequentialSimpleSignalAssignment, since they're the exact same shape and building the mixin made it natural to finish those too.

pyVHDLModel/Common.py
Paebbels: Same WaveformMixin reused here (OthersSelectedWaveform).

pyVHDLModel/Common.py
Paebbels: Extracted a matching ExpressionMixin (Common.py), used by ConditionalExpression, SelectedExpression, and OthersSelectedExpression.

pyVHDLModel/Sequential.py
```diff
@@ -89,40 +93,204 @@ def __init__(
...
+		expression.Parent = self
+
+		self._selectedWaveforms = []
+		for selectedWaveform in selectedWaveforms:
+			self._selectedWaveforms.append(selectedWaveform)
+			selectedWaveform.Parent = self
+
+	@readonly
+	def Expression(self) -> ExpressionUnion:
+		return self._expression
```
Paebbels: Didn't we extract a ExpressionMixin above for signal assignments. Could be reused here too, right?

pyVHDLModel/Sequential.py
```diff
@@ -89,40 +93,204 @@ def __init__(
...
+		super().__init__(label, parent)
+		SignalAssignmentMixin.__init__(self, target)
+
+		self._expression = expression
+		expression.Parent = self
+
+		self._selectedWaveforms = []
+		for selectedWaveform in selectedWaveforms:
+			self._selectedWaveforms.append(selectedWaveform)
+			selectedWaveform.Parent = self
```
Paebbels: This can be extracted into a mixin. See also `ConcurrentSelectedSignalAssignment`.

pyVHDLModel/Sequential.py
```diff
@@ -89,40 +93,204 @@ def __init__(
...
+	) -> None:
+		super().__init__(label, parent)
+		SignalAssignmentMixin.__init__(self, target)
+
+		self._expression = expression
+		expression.Parent = self
+
+	@readonly
+	def Expression(self) -> ExpressionUnion:
+		return self._expression
```
Paebbels: See `ExpressionMixin`.

pyVHDLModel/Sequential.py
Paebbels: Extracted new SelectedWaveformsMixin/SelectedExpressionsMixin (Common.py), mirroring ConditionalWaveformsMixin. Applied to ConcurrentSelectedSignalAssignment, SequentialSelectedVariableAssignment, and SequentialSelectedSignalAssignment.

pyVHDLModel/Sequential.py
Paebbels: Right - now uses ExpressionMixin for the select expression too, alongside the new SelectedExpressionsMixin for the list.

pyVHDLModel/Sequential.py
Paebbels: Fixed - SignalForceAssignment now uses ExpressionMixin.

pyVHDLModel/Common.py
```diff
@@ -152,12 +156,245 @@ class VariableAssignmentMixin(AssignmentMixin, mixin=True):
...
+		WaveformMixin.__init__(self, waveform)
+
+		self._choices = []
+		for choice in choices:
+			self._choices.append(choice)
+			choice.Parent = self
+
+	@readonly
+	def Choices(self) -> List[BaseChoice]:
+		return self._choices
```
Paebbels: We can have a `ChoicesMixin`, right?

This should be checked against case statement and case generate statement, these should use choices too. Either the use already a mixin or they can benefit from such a mixin too.

pyVHDLModel/Common.py
```diff
@@ -152,12 +156,245 @@ class VariableAssignmentMixin(AssignmentMixin, mixin=True):
...
+		ExpressionMixin.__init__(self, expression)
+
+		self._choices = []
+		for choice in choices:
+			self._choices.append(choice)
+			choice.Parent = self
+
+	@readonly
+	def Choices(self) -> List[BaseChoice]:
+		return self._choices
```
Paebbels: `ChoicesMixin` mixin.

pyVHDLModel/Common.py
Paebbels: Good call - checked both. Case statements and case-generate statements had the exact same duplicated pattern already, each marked with a pre-existing TODO ('what about choices?' / 'move to parent or grandparent'). Added ChoicesMixin (Base.py) and applied it to SequentialCase (used by Case/OthersCase) and ConcurrentCase (used by GenerateCase/OthersGenerateCase), in addition to SelectedWaveform/SelectedExpression.

Found a real pre-existing bug while doing this: SequentialCase.__init__ never actually initialized self._choices at all - OthersCase would raise AttributeError the moment .Choices was accessed. Same for OthersGenerateCase via ConcurrentCase. ChoicesMixin fixes this for free since it always initializes the list. Verified directly: OthersCase/OthersGenerateCase.Choices now correctly return [] instead of crashing.

Had to be careful with parameter ordering, since pyGHDL.dom's OthersCase/OthersGenerateCase call super().__init__() positionally without a choices argument - placed the new parameter after the existing ones to avoid breaking that, and verified by running pyGHDL.dom's full test suite (not just this repo's) against the change.

pyVHDLModel/Common.py
Paebbels: Same fix as above - see the reply on the other comment.

PR #139, opened 2026-07-24
Add level-1 construction/property/parent-wiring tests across the model

Paebbels:
## Summary

Systematic level-1 test coverage sweep: instantiate every model entity with required and optional
parameter combinations, verify property accessibility, and check parent-wiring. Driven by
`pytest-cov` to find and prioritize actual gaps rather than guessing. Shared mixins are tested once
via a canonical host rather than duplicated per consuming class; classes with no independent VHDL
construct of their own (`Subprogram`, `Type`/`FullType`/`ScalarType`/`CompositeType`/
`RangedScalarType`, `Allocation`) are tested only through their real concrete subclasses, never
directly - confirmed via grep that pyGHDL.dom never constructs them directly either.

- New test modules: `Base`, `Name`, `Symbol`, `Object`, `Type`, `Expression`, `DesignUnit`,
  `Concurrent`, `Sequential`, `Instantiation`, `Subprogram`; `Interface` and `Configuration` extended.
- Coverage: **71% → 85%** overall, **117 → 361** tests.
- pyGHDL.dom's pyunit suite (123 tests) re-verified clean against every change in this branch.

### Bugs found and fixed while writing these tests

- `Name.OpenName.__init__` — positional-argument mixup crashed any `OpenName(parent=...)` call.
- `Object.DeferredConstant` / `Sequential.LoopControlStatement` — uninitialized slots crashed
  `.ConstantReference` / `.LoopReference` on first access (same shape as the previously-fixed
  `Function.ReturnType` bug).
- **`Type.py` — every concrete type class** (`Subtype`, `EnumeratedType`, `IntegerType`/`RealType`/
  `PhysicalType` via `RangedScalarType`, `ArrayType`, `RecordType`, `ProtectedType`,
  `ProtectedTypeBody`, `AccessType`, `FileType`) forwarded `super().__init__(identifier, parent)`
  positionally, which `BaseType`'s `(identifier, documentation=None, parent=None)` signature silently
  misinterpreted — `parent` landed in the `documentation` slot, and the real `Parent` chain was never
  set. Fixed across all of them, and added a real `documentation` parameter to each (in the same
  position `BaseType` expects it), so every constructor is now a plain, correctly-ordered positional
  forward. This bug was specifically hidden by testing shared behaviour only on the abstract base
  class rather than through each concrete subclass — which is exactly why none of these classes are
  tested directly anymore.
- `Type.FileType.__str__` — said `"...is access..."` (copy-pasted from `AccessType`), now
  `"...is file of..."`.
- `Expression.UnaryExpression` / `Sequential.IndexedChoice` — re-enabled two `Parent`-wiring lines the
  prior gap analysis had already confirmed safe to re-enable.
- `Concurrent.ConcurrentStatementsMixin.IndexStatements` — nested block statements were written to
  `_hierarchy` but `IterateInstantiations` reads `_blocks`, so instantiations nested inside a block
  were invisible to it.
- `Concurrent.IfGenerateStatement`'s `Parent` setter — namespace-naming fallback compared against
  `""` instead of `None`, so it never fired for the common unlabelled-branch case.
- `Subprogram.MethodMixin.__init__` — crashed on its own default (no `protectedType`) case.
- `Interface.GenericGroup` / `ParameterGroup` — missing a base class each (`WithGenericsMixin` /
  `WithParametersMixin`), so construction crashed outright even for an empty list.

### Deliberately left open

Tracked externally for deliberate follow-up rather than fixed inline (each needs a real design
decision, not a one-liner) - happy to share that note if useful:

- **HIGH PRIORITY**: `GenericFunctionInterfaceItem` — confirmed **live**-reachable via pyGHDL.dom's
  actual translation dispatch for VHDL-2008 generic-function clauses, not just a landmine.
- `TernaryExpression`/`WhenElseExpression` — never accepts/sets its operands; separately, `__str__`
  indexes past its `_FORMAT` tuple.
- `TypeConversion.__str__` — no field for the target type at all.
- `GenericGroup`/`PortGroup`/`ParameterGroup.__str__` — crash depending on item mix (singular vs.
  plural identifier shape).
- `QualifiedExpression.Subtyped` — likely-unintentional property name (no call sites anywhere).
- `ConcurrentDeclarationRegionMixin.Functions`/`.Procedures` — silently collide on overloads.
- pyGHDL.dom's own `Type.py` mirror classes now pass a stale positional `None` that lands in the new
  `documentation` slot instead of `parent` - confirmed harmless today (verified against the full
  pyGHDL.dom suite), but worth a follow-up there whenever documentation extraction is wired up.
- Two `Namespace.py` items, set aside per your direction (tested separately).

## Test plan

- [x] `pytest tests/unit` — 361 passed
- [x] pyGHDL.dom pyunit suite (dom/libghdl/lsp) — 123 passed, 5 skipped, 16 xfailed, re-run after every change in this branch

### DIFF
--- a/tests/unit/Name.py
+++ b/tests/unit/Name.py
@@ -1,0 +1,187 @@
+# ==================================================================================================================== #
+#             __     ___   _ ____  _     __  __           _      _                                                     #
+#   _ __  _   \ \   / / | | |  _ \| |   |  \/  | ___   __| | ___| |                                                    #
+#  | '_ \| | | \ \ / /| |_| | | | | |   | |\/| |/ _ \ / _` |/ _ \ |                                                    #
+#  | |_) | |_| |\ V / |  _  | |_| | |___| |  | | (_) | (_| |  __/ |                                                    #
+#  | .__/ \__, | \_/  |_| |_|____/|_____|_|  |_|\___/ \__,_|\___|_|                                                    #
+#  |_|    |___/                                                                                                        #
+# ==================================================================================================================== #
+# Authors:                                                                                                             #
+#   Patrick Lehmann                                                                                                    #
+#                                                                                                                      #
+# License:                                                                                                             #
+# ==================================================================================================================== #
+# Copyright 2026-2026 Patrick Lehmann - Boetzingen, Germany                                                            #
+#                                                                                                                      #
+# Licensed under the Apache License, Version 2.0 (the "License");                                                      #
+# you may not use this file except in compliance with the License.                                                     #
+# You may obtain a copy of the License at                                                                              #
+#                                                                                                                      #
+#   http://www.apache.org/licenses/LICENSE-2.0                                                                         #
+#                                                                                                                      #
+# Unless required by applicable law or agreed to in writing, software                                                  #
+# distributed under the License is distributed on an "AS IS" BASIS,                                                    #
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.                                             #
+# See the License for the specific language governing permissions and                                                  #
+# limitations under the License.                                                                                       #
+#                                                                                                                      #
+# SPDX-License-Identifier: Apache-2.0                                                                                  #
+# ==================================================================================================================== #
+#
+"""Tests for pyVHDLModel.Name."""
+from unittest import TestCase
+
+from pyVHDLModel.Base        import ModelEntity
+from pyVHDLModel.Expression  import IntegerLiteral
+from pyVHDLModel.Association import GenericAssociationItem
+from pyVHDLModel.Name        import (
+	Name, SimpleName, ParenthesisName, IndexedName, SlicedName, SelectedName, AttributeName, AllName, OpenName,
+)
+
+
+if __name__ == "__main__":  # pragma: no cover
+	print("ERROR: you called a testcase declaration file as an executable module.")
+	print("Use: 'python -m unitest <testcase module>'")
+	exit(1)
+
+
+class Names(TestCase):
+	def test_NoPrefixNoParent(self) -> None:
+		name = Name("foo")
+
+		self.assertEqual("foo", name.Identifier)
+		self.assertEqual("foo", name.NormalizedIdentifier)
+		self.assertIsNone(name.Prefix)
+		self.assertFalse(name.HasPrefix)
+		self.assertIs(name, name.Root)
+		self.assertIsNone(name.Parent)
+		self.assertEqual("foo", str(name))
+		self.assertEqual("Name: 'foo'", repr(name))
+
+	def test_IdentifierIsNormalized(self) -> None:
+		name = Name("FOO")
+
+		self.assertEqual("FOO", name.Identifier)
+		self.assertEqual("foo", name.NormalizedIdentifier)
+
+	def test_WithPrefix(self) -> None:
+		"""``Root`` is inherited from the prefix's own root, so for a two-element chain it's simply the
+		prefix itself; see ``test_RootIsTransitiveAcrossAChain`` for a longer chain."""
+		prefix = Name("pkg")
+		name = Name("member", prefix)
+
+		self.assertIs(prefix, name.Prefix)
+		self.assertTrue(name.HasPrefix)
+		self.assertIs(prefix, name.Root)
+
+	def test_RootIsTransitiveAcrossAChain(self) -> None:
+		root = Name("a")
+		middle = Name("b", root)
+		leaf = Name("c", middle)
+
+		self.assertIs(root, middle.Root)
+		self.assertIs(root, leaf.Root)
+		self.assertIs(middle, leaf.Prefix)
+
+	def test_WithParent(self) -> None:
+		parent = ModelEntity()
+		name = Name("foo", parent=parent)
+
+		self.assertIs(parent, name.Parent)
+
+
+class SimpleNames(TestCase):
+	def test_Construction(self) -> None:
+		name = SimpleName("sig")
+
+		self.assertEqual("sig", name.Identifier)
+		self.assertFalse(name.HasPrefix)
+
+
+class ParenthesisNames(TestCase):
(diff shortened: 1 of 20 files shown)


tests/unit/Name.py
```diff
@@ -0,0 +1,187 @@
...
+)
+
+
+if __name__ == "__main__":  # pragma: no cover
+	print("ERROR: you called a testcase declaration file as an executable module.")
+	print("Use: 'python -m unitest <testcase module>'")
+	exit(1)
+
+
+class NameClass(TestCase):
```
Paebbels: Why the postfix `*Class`? can you use `Names`?

tests/unit/Object.py
```diff
@@ -0,0 +1,180 @@
...
+		self.assertIs(default, signal.DefaultExpression)
+		self.assertIs(signal, default.Parent)
+
+	def test_WithoutDefaultExpression(self) -> None:
+		signal = Signal(["s"], _subtype())
+
+		self.assertIsNone(signal.DefaultExpression)
+
+
+class ConstantClass(TestCase):
```
Paebbels: maybe name it `Constants`.

tests/unit/Sequential.py
```diff
@@ -0,0 +1,274 @@
...
+
+		self.assertIsNone(statement.SensitivityList)
+		self.assertIsNone(statement.Condition)
+		self.assertIsNone(statement.Timeout)
+
+	def test_Full(self) -> None:
+		"""``wait on clk until cond for 10 ns;``"""
+		sensitivitySignal = _signalTarget("clk")
+		condition = IntegerLiteral(1)
+		timeout = IntegerLiteral(10)
```
Paebbels: a timeout must be a pysical literal of type time or an expression calculating a time. Use `10 ns` as example.

tests/unit/Sequential.py
```diff
@@ -0,0 +1,274 @@
...
+		"""``wait;``"""
+		statement = WaitStatement()
+
+		self.assertIsNone(statement.SensitivityList)
+		self.assertIsNone(statement.Condition)
+		self.assertIsNone(statement.Timeout)
+
+	def test_Full(self) -> None:
+		"""``wait on clk until cond for 10 ns;``"""
+		sensitivitySignal = _signalTarget("clk")
```
Paebbels: don't use clk or rst, prefer clock and reset.

tests/unit/Sequential.py
```diff
@@ -0,0 +1,274 @@
...
+		condition = IntegerLiteral(1)
+		timeout = IntegerLiteral(10)
+		statement = WaitStatement([sensitivitySignal], condition, timeout)
+
+		self.assertEqual(1, len(statement.SensitivityList))
+		self.assertIs(sensitivitySignal, statement.SensitivityList[0])
+		self.assertIs(statement, sensitivitySignal.Parent)
+		self.assertIs(condition, statement.Condition)
+		self.assertIs(timeout, statement.Timeout)
```
Paebbels, line 274: Don't you need to check condition's parent relationship?

tests/unit/Subprogram.py
```diff
@@ -0,0 +1,142 @@
...
+	return SimpleSubtypeSymbol(SimpleName(name))
+
+
+class SubprogramClass(TestCase):
+	"""``Subprogram`` itself is instantiable directly (unlike most other "base" classes in this
+	model) - ``Procedure``/``Function`` only add ``isPure``/``ReturnType`` on top, so its own
+	declared-items/statements wiring is tested once here rather than duplicated in both."""
+
+	def test_Minimal(self) -> None:
+		subprogram = Subprogram("sub", isPure=True)
```
Paebbels: subprogram is abstract (not marked so). Users should instantiate a subprogram directly. please move tests into derived classes and don't instantiate "abstract classes" (classes that should be used for inheritance only) I'll implement an abstract class feature later in pyTooling (you can check pyTooling.Metaclass.ExtendedType if it already supports an @abstract or @abstractclass as a class decorator)

tests/unit/Name.py
Paebbels: Renamed to `Names` (and did the same drop-the-`Class`-suffix rename across every new test file, not just this one).

tests/unit/Object.py
Paebbels: Renamed to `Constants`.

tests/unit/Sequential.py
Paebbels: Fixed - now uses `PhysicalIntegerLiteral(10, "ns")` for the timeout.

tests/unit/Sequential.py
Paebbels: Renamed to `clock`.

tests/unit/Sequential.py
Paebbels: Good catch, added `self.assertIs(statement, condition.Parent)`.

tests/unit/Subprogram.py
Paebbels: Agreed - moved the shared declared-items/statements/generic-items/parameter-items test onto `Procedure` instead (the simpler of the two real subclasses), and dropped direct `Subprogram(...)` construction entirely from this file.

Paebbels:
History rewritten: squashed everything into a single commit and force-pushed - `Findings.md` (my working notes file, per your request) never appears in the history at all now, since it was added and removed within the original set of commits and nets to zero in the squash.

The branch now has exactly one commit: **"Add level-1 construction/property/parent-wiring tests across the model"**. PR description above is updated to match.

Everything from the review round is included in that single commit:
- Dropped the `*Class` test-name suffix everywhere (`Names`, `Constants`, etc.)
- `Subprogram`/`Type`/`Allocation` are no longer instantiated directly in tests - only through concrete subclasses
- The `WaitStatement` test fixes (clock, physical literal, condition.Parent)
- The `documentation`-parameter fix for every `Type.py` leaf class, verified it can actually be set (not just that it stays `None`)

361 tests passing, pyGHDL.dom's 123-test suite still clean.

PR #140, opened 2026-07-24
Three small cleanups from the test-strategy findings

Paebbels:
## Summary

Three small cleanups from the recent test-strategy work (#139) - items 1-3 from the follow-up TODO
list. Design decisions for each were discussed and confirmed before implementing.

- **`QualifiedExpression.Subtyped` -> `Subtype`**: every other symmetric case in the model
  (`SubtypeAllocation.Subtype`, `Subtype.Type`, `ConstrainedScalarSubtypeSymbol`, ...) reads
  `.Subtype`; this property had zero call sites anywhere (grepped both repos), so the rename is
  no-risk.
- **`DesignUnitWithContextMixin`** (`Context.__init__`) raised a bare `VHDLModelException()` with no
  message when a reference is neither a `LibraryClause`, `UseClause`, nor `ContextReference`. Now
  includes the offending reference in the message.
- **`ConcurrentStatementsMixin`'s `_blocks`/`_generates`/`_hierarchy` bookkeeping** was a
  partially-implemented, redundant three-dict setup: `_hierarchy` was write-only
  (`IndexStatements` wrote nested block statements into it, but `IterateInstantiations` only ever
  read the separate `_blocks` dict, which got the same writes), while `_generates` duplicated the
  same "just filter by kind" role for generate statements. Unified all of it into a single,
  source-order `_hierarchy` dict (indexing both blocks and generates, in declaration order - the
  same order used to build the elaborated design hierarchy) and derive the blocks-only/
  generates-only views by filtering on `isinstance` where needed (currently just
  `IterateInstantiations`), rather than maintaining three overlapping dicts by hand.

## Test plan

- [x] `pytest tests/unit` - 356 passed
- [x] pyGHDL.dom pyunit suite - 123 passed, 5 skipped, 16 xfailed (unaffected, re-verified)

### DIFF
--- a/pyVHDLModel/Concurrent.py
+++ b/pyVHDLModel/Concurrent.py
@@ -84,16 +84,12 @@ class ConcurrentStatementsMixin(metaclass=ExtendedType, mixin=True):
 	_statements:     List[ConcurrentStatement]
 
 	_instantiations: Dict[str, 'Instantiation']  # TODO: add another instantiation class level for entity/configuration/component inst.
-	_blocks:         Dict[str, 'ConcurrentBlockStatement']
-	_generates:      Dict[str, 'GenerateStatement']
-	_hierarchy:      Dict[str, Union['ConcurrentBlockStatement', 'GenerateStatement']]
+	_hierarchy:      Dict[str, Union['ConcurrentBlockStatement', 'GenerateStatement']]  #: All elements creating a hierarchy level (blocks and generates), in declaration order.
 
 	def __init__(self, statements: Nullable[Iterable[ConcurrentStatement]] = None) -> None:
 		self._statements = []
 
 		self._instantiations = {}
-		self._blocks = {}
-		self._generates = {}
 		self._hierarchy = {}
 
 		if statements is not None:
@@ -109,11 +105,8 @@ class ConcurrentStatementsMixin(metaclass=ExtendedType, mixin=True):
 		for instance in self._instantiations.values():
 			yield instance
 
-		for block in self._blocks.values():
-			yield from block.IterateInstantiations()
-
-		for generate in self._generates.values():
-			yield from generate.IterateInstantiations()
+		for element in self._hierarchy.values():
+			yield from element.IterateInstantiations()
 
 	# TODO: move into _init__
 	def IndexStatements(self) -> None:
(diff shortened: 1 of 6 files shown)


pyVHDLModel/Concurrent.py
```diff
@@ -109,22 +105,21 @@ def IterateInstantiations(self) -> Generator['Instantiation', None, None]:
...
-		for block in self._blocks.values():
-			yield from block.IterateInstantiations()
-
-		for generate in self._generates.values():
-			yield from generate.IterateInstantiations()
+		for element in self._hierarchy.values():
+			if isinstance(element, ConcurrentBlockStatement):
+				yield from element.IterateInstantiations()
+			elif isinstance(element, GenerateStatement):
+				yield from element.IterateInstantiations()
```
Paebbels: Why the isinstance, when it calls the same method?

pyVHDLModel/Concurrent.py
```diff
@@ -84,16 +84,12 @@ class ConcurrentStatementsMixin(metaclass=ExtendedType, mixin=True):
 	_statements:     List[ConcurrentStatement]
 
 	_instantiations: Dict[str, 'Instantiation']  # TODO: add another instantiation class level for entity/configuration/component inst.
-	_blocks:         Dict[str, 'ConcurrentBlockStatement']
-	_generates:      Dict[str, 'GenerateStatement']
-	_hierarchy:      Dict[str, Union['ConcurrentBlockStatement', 'GenerateStatement']]
+	_hierarchy:      Dict[str, Union['ConcurrentBlockStatement', 'GenerateStatement']]  #: All elements creating a hierarchy level (blocks and generates), in source/declaration order.
```
Paebbels:
```suggestion
	_hierarchy:      Dict[str, Union['ConcurrentBlockStatement', 'GenerateStatement']]  #: All elements creating a hierarchy level (blocks and generates), in declaration order.
```

tests/unit/Concurrent.py
```diff
@@ -371,6 +371,8 @@ class ConcurrentStatementsMixinIndexing(TestCase):
 	``ConcurrentStatementsMixin`` host - see tests/unit/DesignUnit.py)."""
 
 	def test_IndexStatements_BucketsByKind(self) -> None:
+		"""Blocks and generates both land in the single, unified ``_hierarchy`` dict (in source
+		order), not separate per-kind dicts - see Findings.md for the design note."""
```
Paebbels: do not reference our private discussion files like findings.md. This doesn't belong into the repository. You need to amend or squash your commit.

pyVHDLModel/Concurrent.py
Paebbels: Good catch - both branches called the identical `element.IterateInstantiations()`, so the isinstance check was pointless. Simplified to a plain `for element in self._hierarchy.values(): yield from element.IterateInstantiations()`.

pyVHDLModel/Concurrent.py
Paebbels: Applied, thanks.

tests/unit/Concurrent.py
Paebbels: Fixed - removed, and swept both PRs (#140 and #141) for the same mistake. Turns out it's bigger than this one spot: PR #139 (already merged into dev) has 7 more of these baked into test docstrings. I'll open a follow-up PR to clean those out of dev separately, since that history is already merged.

pyVHDLModel/Expression.py
```diff
@@ -652,7 +652,7 @@ def Operand(self):
 		return self._operand
 
 	@property
```
Paebbels:
```suggestion
	@readonly
```

PR #141, opened 2026-07-24
Fix four expression/interface/region gaps from the test-strategy findings

Paebbels:
## Summary

Items 4-7 from the follow-up TODO list after #139/#140. Design decisions for each were discussed and
confirmed (some revised) before finalizing - see the PR conversation/commit message for the full
back-and-forth. Updated after review: `TypeConversion`/`WhenElseExpression` now use `SubtypeSymbol`
and `@readonly` per feedback (see below).

- **`Expression.TernaryExpression`** - the constructor never accepted or set its three operands at
  all, and `__str__` indexed past the end of its `_FORMAT` tuple. Fixed: constructor now takes
  `(firstOperand, secondOperand, thirdOperand)`, wiring `Parent` for each. Deliberately exposes **no**
  public `FirstOperand`/`SecondOperand`/`ThirdOperand` - unlike `Unary`/`BinaryExpression` (where the
  generic name already *is* the domain name for every consumer), a ternary's three operands mean
  different things per concrete expression, so `WhenElseExpression` (`thenValue when condition else
  elseValue`) gets its own `__init__` and its own `@readonly` `ThenValue`/`Condition`/`ElseValue`
  properties instead of duplicating a generic and a specific name for the same value. `_FORMAT` is
  marked `# FIXME: needs ClassVar[...] when pyTooling gets fixed.`

- **`Expression.TypeConversion`** had no field for the target type at all, so `str()` always crashed.
  Added `_targetSubtype`/`TargetSubtype` (named more specifically than just `Subtype`, since
  `TypeConversion` already has its own `operand` to disambiguate from), typed as `SubtypeSymbol`
  (added to `pyVHDLModel/Symbol.py`, mirroring `PackageSymbol`) rather than the generic `Symbol` base,
  and exposed via `@readonly`. pyGHDL.dom never constructs this class today (confirmed via grep) -
  the translation-side follow-up is tracked separately, not attempted here.

- **`Interface.GenericGroup`/`PortGroup`/`ParameterGroup.__str__`** crashed depending on item mix
  (interface items come in two incompatible identifier shapes - singular `NamedEntityMixin` vs.
  plural `MultipleNamedEntityMixin`, since one declaration can name several objects:
  `port (p1, p2 : in bit);`). Fixed narrowly via a shared `_identifiersOf()` helper
  (`isinstance`-based) that flattens either shape into a plain tuple of names. A bigger redesign (give
  `NamedEntityMixin`/`MultipleNamedEntityMixin` a real `__str__` so groups can just delegate to
  `str(item)`) was identified but deliberately deferred to its own branch - touches two mixins
  inherited by dozens of unrelated classes, so it needs its own review pass.

- **`Regions.ConcurrentDeclarationRegionMixin.Functions`/`.Procedures`** silently collided on
  overloaded subprograms sharing a name (flat dict, second overwrites first). A textual-signature
  approach was tried and rejected as unreliable (aliased/case-differing subtype names would be
  misjudged as distinct, and symbol resolution happening after indexing could change the comparison
  basis). Simplified instead to `Dict[str, List[Function]]`/`Dict[str, List[Procedure]]` -
  collision-avoidance only, marked with `# FIXME` at both call sites: overloads are collected, not
  actually resolved by signature. Real overload resolution needs type-compatibility matching, not
  string comparison - tracked as a substantial follow-up feature.

This branch was forked from `dev` before #140 merged, so it doesn't include that PR's changes
(`QualifiedExpression.Subtyped` rename, `DesignUnitWithContextMixin` exception message,
`_blocks`/`_generates`/`_hierarchy` cleanup) - independent, non-overlapping work.

## Test plan

- [x] `pytest tests/unit` - 355 passed
- [x] pyGHDL.dom pyunit suite - 123 passed, 5 skipped, 16 xfailed (none of the four changed
      constructors/behaviours are currently reachable from its translation dispatch, grepped - no
      cross-repo fallout)

### DIFF
--- a/pyVHDLModel/Expression.py
+++ b/pyVHDLModel/Expression.py
@@ -40,7 +40,7 @@ from typing               import Tuple, List, Iterable, Union, ClassVar, Optiona
 from pyTooling.Decorators import export, readonly
 
 from pyVHDLModel.Base     import ModelEntity, Direction, Range
-from pyVHDLModel.Symbol   import Symbol
+from pyVHDLModel.Symbol   import Symbol, SubtypeSymbol
 
 
 ExpressionUnion = Union[
@@ -364,7 +364,26 @@ class AbsoluteExpression(UnaryExpression):
 
 @export
 class TypeConversion(UnaryExpression):
-	pass
+	"""``natural(x)`` - a type conversion. Unlike every other :class:`UnaryExpression` subclass, its
+	"operator" is the target type name itself, not a fixed string, so it doesn't participate in the
+	shared ``_FORMAT``-based :meth:`UnaryExpression.__str__` at all - it carries its own
+	:attr:`_targetSubtype` (mirroring :class:`QualifiedExpression`'s ``_subtype``) and overrides
+	``__str__`` accordingly."""
+
+	_targetSubtype: SubtypeSymbol
+
+	def __init__(self, targetSubtype: SubtypeSymbol, operand: ExpressionUnion, parent: Nullable[ModelEntity] = None) -> None:
+		super().__init__(operand, parent)
+
+		self._targetSubtype = targetSubtype
+		targetSubtype.Parent = self
+
+	@readonly
+	def TargetSubtype(self) -> SubtypeSymbol:
+		return self._targetSubtype
+
+	def __str__(self) -> str:
+		return f"{self._targetSubtype!s}({self._operand!s})"
 
 
 @export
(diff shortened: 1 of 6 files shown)


pyVHDLModel/Expression.py
```diff
@@ -364,7 +364,26 @@ class AbsoluteExpression(UnaryExpression):
...
 @export
 class TypeConversion(UnaryExpression):
-	pass
+	"""``natural(x)`` - a type conversion. Unlike every other :class:`UnaryExpression` subclass, its
+	"operator" is the target type name itself, not a fixed string, so it doesn't participate in the
+	shared ``_FORMAT``-based :meth:`UnaryExpression.__str__` at all - it carries its own
+	:attr:`_targetSubtype` (mirroring :class:`QualifiedExpression`'s ``_subtype``) and overrides
+	``__str__`` accordingly."""
+
+	_targetSubtype: Symbol
```
Paebbels: Don't we have a more specific Symbol like a SubtypeSymbol? If not, we need one. See PackageSymbol as a reference.

pyVHDLModel/Expression.py
```diff
@@ -661,29 +680,41 @@ def __str__(self) -> str:
...
+	:class:`UnaryExpression`/:class:`BinaryExpression` (where "the operand"/"left operand"/"right
+	operand" are already the natural domain names for every consumer), a ternary's three operands
+	play a different semantic role depending on the concrete expression - e.g.
+	:class:`WhenElseExpression`'s are really "then value"/"condition"/"else value". Each concrete
+	subclass is expected to expose its own, appropriately-named properties over the shared
+	``_firstOperand``/``_secondOperand``/``_thirdOperand`` fields, rather than duplicating a
+	generic and a specific name for the same value.
+	"""
 
 	_FORMAT: Tuple[str, str, str, str]
```
Paebbels:
```suggestion
	_FORMAT: Tuple[str, str, str, str]  # FIXME: needs ClassVar[...] when pyTooling gets fixed.
```

pyVHDLModel/Expression.py
```diff
@@ -693,14 +724,41 @@ def __str__(self) -> str:
...
+	def __init__(
+		self,
+		thenValue: ExpressionUnion,
+		condition: ExpressionUnion,
+		elseValue: ExpressionUnion,
+		parent: Nullable[ModelEntity] = None
+	) -> None:
+		super().__init__(thenValue, condition, elseValue, parent)
+
+	@property
```
Paebbels: `@readonly`

pyVHDLModel/Expression.py
Paebbels: Fixed - added SubtypeSymbol (matching PackageSymbol's pattern) and changed TargetSubtype/_targetSubtype to use it instead of the generic Symbol.

pyVHDLModel/Expression.py
Paebbels: Fixed - added the suggested comment.

pyVHDLModel/Expression.py
Paebbels: Fixed - changed to @readonly. Also caught TypeConversion.TargetSubtype, which had the same bare @property mistake but wasn't flagged here.

PR #143, opened 2026-07-24
Fix GenericFunctionInterfaceItem crash on real VHDL-2008 generic function clauses

Paebbels:
## Summary

Item 8 of the follow-up TODO list, and the one confirmed-live cross-repo bug: pyGHDL.dom's
`_Translate.py` calls `GenericFunctionInterfaceItem.parse(generic)` while translating any
`generic (function f return t);` clause, and that `.parse()` called
`super().__init__(identifier, documentation, parent)` - but `GenericFunctionInterfaceItem` had no
`returnType` parameter of its own, so the call misaligned against `Function.__init__`'s real
signature `(identifier, returnType, isPure, ...)`: `documentation` silently landed in the
`returnType` slot, and `Function.__init__` unconditionally does `returnType.Parent = self`,
crashing immediately (`AttributeError`) on anything but a real `Symbol` - including the plain,
no-docs case, since `None` has no `Parent` either.

**Fixed** by adding a real `returnType: SubtypeSymbol` parameter to
`GenericFunctionInterfaceItem.__init__`, forwarding `documentation`/`parent` as keyword
arguments (not positionally) to `Function.__init__` to avoid the exact same mismatch shape
recurring. The companion pyGHDL.dom-side fix (reading the return type mark off the IIR node) is a
separate commit in that repo.

**Also found and fixed while here:**

- `GenericProcedureInterfaceItem` had the identical positional-forwarding bug against
  `Procedure.__init__` - never confirmed live (no test ever passed a non-`None` `documentation`),
  but fixed alongside for the same reason.
- A genuine **diamond inheritance bug** affecting all 10 `Generic*`/`Parameter*InterfaceItem`
  classes: each calls its own mixin's `__init__(self)` with no arguments *after* the primary base
  already set `documentation` correctly. `InterfaceItemMixin` (the base of
  `GenericInterfaceItemMixin`/`ParameterInterfaceItemMixin`) also inherited
  `DocumentedEntityMixin`, so every one of these classes reached `DocumentedEntityMixin` via two
  independent paths - the domain base (`Constant`/`Type`/`Function`/`Procedure`/`Signal`/
  `Variable`/`File`/`InterfacePackage`) and the interface-item mixin. Python's C3 linearization
  deduplicates the diamond to a single MRO slot, but this codebase's constructors call every base's
  `__init__` explicitly by name rather than cooperatively via `super()`, so the mixin's
  argument-less call re-entered `DocumentedEntityMixin.__init__` a second time with
  `documentation=None`, silently overwriting the value the primary base had already set correctly.
  **Fixed at the class-graph level**: removed `DocumentedEntityMixin` from `InterfaceItemMixin`'s
  bases instead of removing the redundant calls - every concrete interface item is already a
  documentable entity via its primary base, so the mixin never needed to carry documentation of its
  own. The `Generic*`/`ParameterInterfaceItemMixin.__init__(self)` calls stay in place (now
  genuinely inert), preserving the existing constructor shape everywhere else in the file.

## Test plan

- [x] `pytest tests/unit` - 357 passed
- [x] pyGHDL.dom pyunit suite - 125 passed (up from 123; two new regression tests added there),
      5 skipped, 16 xfailed, after the companion translation-side fix

### DIFF
--- a/pyVHDLModel/Interface.py
+++ b/pyVHDLModel/Interface.py
@@ -182,11 +182,20 @@ class ModeViewDeclaration(ModelEntity, NamedEntityMixin, DocumentedEntityMixin):
 
 
 @export
-class InterfaceItemMixin(DocumentedEntityMixin, mixin=True):
-	"""An ``InterfaceItem`` is a base-class for all mixin-classes for all interface items."""
+class InterfaceItemMixin(metaclass=ExtendedType, mixin=True):
+	"""
+	A base-class for all mixin-classes for all interface items.
 
-	def __init__(self, documentation: Nullable[str] = None) -> None:
-		super().__init__(documentation)
+	Doesn't inherit :class:`~pyVHDLModel.Base.DocumentedEntityMixin` itself: every concrete interface
+	item is already a documentable language entity in its own right (a constant, signal, variable,
+	file, type, subprogram or package) via its *other*, primary base, which provides
+	:class:`DocumentedEntityMixin` on its own. Adding it here too would create a diamond back to the
+	same mixin, reachable through two independent paths - harmless for Python's MRO (which
+	deduplicates it to a single slot), but a landmine for this codebase's constructor convention of
+	calling each base's ``__init__`` explicitly instead of cooperatively through ``super()``: the
+	second, argument-less call down this path would silently reset ``documentation`` back to
+	``None`` after the primary base had already set it correctly.
+	"""
 
 
 @export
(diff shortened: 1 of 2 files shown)


pyVHDLModel/Interface.py
```diff
@@ -182,11 +182,20 @@ def Elements(self) -> List[ModeViewElement]:
...
+	Doesn't inherit :class:`~pyVHDLModel.Base.DocumentedEntityMixin` itself: every concrete interface
+	item is already a documentable language entity in its own right (a constant, signal, variable,
+	file, type, subprogram or package) via its *other*, primary base, which provides
+	:class:`DocumentedEntityMixin` on its own. Adding it here too would create a diamond back to the
+	same mixin, reachable through two independent paths - harmless for Python's MRO (which
+	deduplicates it to a single slot), but a landmine for this codebase's constructor convention of
+	calling each base's ``__init__`` explicitly instead of cooperatively through ``super()``: the
+	second, argument-less call down this path would silently reset ``documentation`` back to
+	``None`` after the primary base had already set it correctly.
+	"""
```
Paebbels, line 188:
```suggestion
	A base-class for all mixin-classes for all interface items.
	"""
```

pyVHDLModel/Interface.py
```diff
@@ -253,14 +262,20 @@ class GenericSubprogramInterfaceItem(GenericInterfaceItemMixin):
 @export
 class GenericProcedureInterfaceItem(Procedure, GenericInterfaceItemMixin):
 	def __init__(self, identifier: str, documentation: Nullable[str] = None, parent: Nullable[ModelEntity] = None) -> None:
-		super().__init__(identifier, documentation, parent)
+		super().__init__(identifier, documentation=documentation, parent=parent)
```
```suggestion
		super().__init__(identifier, documentation, parent)
```

pyVHDLModel/Interface.py
```diff
@@ -253,14 +262,20 @@ class GenericSubprogramInterfaceItem(GenericInterfaceItemMixin):
...
-	def __init__(self, identifier: str, documentation: Nullable[str] = None, parent: Nullable[ModelEntity] = None) -> None:
-		super().__init__(identifier, documentation, parent)
+	def __init__(
+		self,
+		identifier:    str,
+		returnType:    SubtypeSymbol,
+		documentation: Nullable[str] = None,
+		parent:        Nullable[ModelEntity] = None
+	) -> None:
+		super().__init__(identifier, returnType, documentation=documentation, parent=parent)
```
```suggestion
		super().__init__(identifier, returnType, documentation, parent)
```

PR #145, opened 2026-07-26
Split Range into SimpleRange and RangeFromName

Paebbels:
Follows up on the review discussion in [Paebbels/ghdl#14](https://github.com/Paebbels/ghdl/pull/14#discussion_r3651299033): a range denoted by a name had nowhere to go in the model, so pyGHDL.dom was passing a bare `Name` into range-typed slots. Anything reading `.LeftBound`/`.Direction` off such a field would break.

## Grammar

Checked against the [machine-readable VHDL-2019 EBNF](https://www.sigasi.com/tech/vhdl2017.ebnf/):

```
range              ::= range_attribute_name | simple_range | range_expression
simple_range       ::= simple_expression direction simple_expression
discrete_range     ::= discrete_subtype_indication | range
subtype_indication ::= [ resolution_indication ] type_mark [ constraint ]
```

The model had only the `simple_range` form.

## Change

`Range` becomes the base-class:

- **`SimpleRange`** carries the previous bounds/direction implementation, named after the grammar's `simple_range` rule.
- **`RangeFromName`** wraps a `Symbol`, so bounds are inferred once that symbol is resolved.

`RangeFromName` deliberately covers both the range-attribute form (`vector'range`) and the discrete subtype indication form (`bit`, `integer range 0 to 7`). Strictly, the LRM puts a subtype indication one level up (`discrete_range ::= discrete_subtype_indication | range`), but a parser only ever sees "a name, optionally with a range constraint", and keeping one uniform type at every usage site is worth more here than mirroring the rule split. This is stated in the class docstring.

A constrained subtype indication no longer loses its type mark — it's carried by a `ConstrainedScalarSubtypeSymbol`, which holds the type mark and the range constraint together.

**`RangeAttributeSymbol`** is new. `PossibleReference.RangeAttribute` was declared in the enumeration but no symbol class ever used it; `RangeFromName` needs one for the attribute form.

## `RangedScalarType` migration

`integer_type_definition ::= range_constraint`, and `range_constraint ::= range range` — so an integer/floating/physical type definition is *always* a range, never a subtype indication. Its `Union[Range, Name]` existed only because a range attribute had no `Range` representation; now it does, so the field becomes a plain `Range`.

Its `_leftBound`/`_rightBound` fields are also removed: they were declared but never assigned and had no properties, so with `slots=True` they didn't exist at runtime.

## Testing

361 passed (357 before, +4 for the new classes), covering construction, parent wiring, `__str__`, and that a constrained subtype indication keeps both its type mark and its range constraint.

Companion pyGHDL.dom change in [Paebbels/ghdl#14](https://github.com/Paebbels/ghdl/pull/14) builds these classes from real IIR nodes; it needs this PR merged and released first.

### DIFF
--- a/pyVHDLModel/Base.py
+++ b/pyVHDLModel/Base.py
@@ -481,11 +481,35 @@ class ChoicesMixin(metaclass=ExtendedType, mixin=True):
 
 @export
 class Range(ModelEntity):
+	"""
+	Base-class for all ranges.
+
+	VHDL's ``range`` rule offers a range denoted by a name (:class:`RangeFromName`) as well as a range
+	given by explicit bounds (:class:`SimpleRange`).
+	"""
+
+
+@export
+class SimpleRange(Range):
+	"""
+	A range with both bounds given as expressions, e.g. ``0 to 7``.
+
+	Named after the ``simple_range`` rule in VHDL's grammar.
+	"""
+
 	_leftBound:  ExpressionUnion
 	_rightBound: ExpressionUnion
 	_direction:  Direction
 
 	def __init__(self, leftBound: ExpressionUnion, rightBound: ExpressionUnion, direction: Direction, parent: Nullable[ModelEntity] = None) -> None:
+		"""
+		Initialize a simple range.
+
+		:param leftBound:  The range's left bound.
+		:param rightBound: The range's right bound.
+		:param direction:  The range's direction (``to`` or ``downto``).
+		:param parent:     The parent model entity.
+		"""
 		super().__init__(parent)
 
 		self._leftBound = leftBound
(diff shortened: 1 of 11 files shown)


pyVHDLModel/Base.py
```diff
@@ -481,11 +481,35 @@ def Choices(self) -> List[BaseChoice]:
...
+	given by explicit bounds (:class:`SimpleRange`).
+	"""
+
+
+@export
+class SimpleRange(Range):
+	"""
+	A range with both bounds given as expressions, e.g. ``0 to 7``.
+
+	Named after the ``simple_range`` rule in VHDL's grammar.
```
Paebbels:
```suggestion
```

pyVHDLModel/Base.py
```diff
@@ -496,22 +520,65 @@ def __init__(self, leftBound: ExpressionUnion, rightBound: ExpressionUnion, dire
...
+@export
+class RangeFromName(Range):
+	"""
+	A range denoted by a name, so its bounds are inferred from whatever that name references.
+
+	Two forms reach this class, because a parser can't tell them apart beyond "a name, optionally with a
+	range constraint":
+
+	* a range attribute like ``vector'range``, and
+	* a discrete subtype indication like ``bit`` or ``integer range 0 to 7``.
```
Paebbels: Embedd this in a ReST note `.. note::`.

pyVHDLModel/Type.py
```diff
@@ -125,16 +124,23 @@ class ScalarType(FullType):
 class RangedScalarType(ScalarType):
 	"""A ``RangedScalarType`` is a base-class for all scalar types with a range."""
 
-	_range:      Union[Range, Name]
-	_leftBound:  ExpressionUnion
-	_rightBound: ExpressionUnion
```
Paebbels, line 130: These where unused? Then it's ok to remove them.

tests/unit/Base.py
```diff
@@ -222,10 +224,43 @@ def test_Construction(self) -> None:
...
+
+		self.assertIsInstance(range_, Range)
+
+
+class RangesFromName(TestCase):
+	"""``RangeFromName`` (``vector'range``, ``bit``) - a range whose bounds come from a referenced symbol."""
+
+	def test_Construction(self) -> None:
+		symbol = SimpleSubtypeSymbol(SimpleName("bit"))
+		range_ = RangeFromName(symbol)
```
Paebbels: In other places I used `rng` for range.

pyVHDLModel/Base.py
Paebbels: Removed — `SimpleRange`'s docstring is now just the one-line summary.

pyVHDLModel/Base.py
Paebbels: Done. The "two forms" explanation is now a `.. note::`, with the main text above it describing what the class holds. I also spelled out the grammar deviation inside the note, so it's explicit rather than implied:

```
.. note::

   Two forms reach this class, because a parser can't tell them apart beyond "a name, optionally with
   a range constraint":

   * a range attribute like ``vector'range``, and
   * a discrete subtype indication like ``bit`` or ``integer range 0 to 7``.

   VHDL's grammar puts the latter one level up (``discrete_range ::= discrete_subtype_indication |
   range``), so representing both as a range deviates from the rule split deliberately.
```

pyVHDLModel/Type.py
Paebbels: Yes — genuinely unused. `_leftBound`/`_rightBound` were declared as annotations on `RangedScalarType` but never assigned in `__init__`, and no property exposed them. Since `ExtendedType` derives `__slots__` from those annotations and the class uses `slots=True`, the attributes did not exist at runtime at all — reading either one raised `AttributeError`. Nothing in the repo referenced them.

They looked like an unfinished first attempt at the same "infer bounds from the range" idea that `RangeFromName` now handles.

tests/unit/Base.py
Paebbels: Renamed. I applied it to the whole file rather than just the new tests, so `tests/unit/Base.py` is now internally consistent with `Sequential.py`/`Concurrent.py` — the two pre-existing `range_` uses in `Ranges` were renamed too. Say the word if you would rather I had left those alone.

PR #147, opened 2026-07-26
Namespace tests, ExtendedKeyError chaining, and the Instantiation test package

Paebbels:
Implements the testing strategy agreed on 2026-07-26, and closes the two `Namespace` findings that were waiting on it.

Three reviewable commits: a pure move, then the source fix with its tests, then the relocations.

## 1. `tests/unit/Instantiation/` package (rename-only)

`tests/unit` gains the package level that names the *capability under test*, so packages testing combinations of entities can sit beside the per-class construction tests. Pure `git mv` — **361 passed before and after**, no test content touched.

Two renames were needed:

| Was | Now | Why |
|---|---|---|
| `Instantiation.py` | `GenericInstantiation.py` | A module can't live inside a package of the same name. Still covers VHDL-2008 generic subprogram/package instantiations. |
| `Instantiate.py` | `Model.py` | The package now carries the "instantiation" meaning; the module covers the model's top-level composition (names, symbols, design, library, document). |

`Analyze.py` and `Hierarchy.py` deliberately **stay at `tests/unit`** — they test dependency graphs, linking and indexing rather than construction, and belong in `Indexing/` and `DependencyGraph/` in the next step.

## 2. `ExtendedKeyError` chaining + copy-pasted messages

`FindComponent` accumulates the searched namespaces across the parent walk and reports them all (`"... not found in: architecture, entity, library."`). `FindSubtype` and `FindObject` still raised a plain `KeyError` naming only the innermost namespace, so a failed lookup gave no clue which scopes were searched. Both now follow the same pattern, including `raise ... from ex`.

Two `FindObject` messages were copy-pasted from `FindSubtype` and named the wrong thing:

- wrong-kind branch: `"it is not a type or subtype"` → `"it is not an object"`
- not-found branch: `"Subtype 'x' not found"` → `"Object 'x' not found"`

New `tests/unit/Namespace/Namespace.py` covers `Namespace` standalone — construction, parent/sub-namespace wiring, and all three `Find*` methods across found / found-via-normalized-identifier / wrong-kind / not-expected / not-found-locally / not-found-anywhere, asserting `key`, `searchedNamespaces` ordering, the message, and the `__cause__` chain.

**Five of those assertions were confirmed to fail without the source fix** (and pass with it), so they're genuine regression tests rather than descriptions of current behaviour.

`Namespace` has no public insert method — `_elements` is populated by `IndexDeclaredItems` — so the tests seed `_elements` directly, keyed by *normalized* identifier exactly as the `Find*` methods look it up. That's stated in the module docstring.

## 3. Namespace-usage assertions relocated

Namespace wiring isn't a variant of whichever feature happens to own it, so it moves out of the instantiation classes into `Namespace/Statements.py` (blocks, all three generate kinds) and `Namespace/DesignUnits.py` (`IndexDeclaredItems`). Bodies unchanged; four method names shortened because the module now supplies the context the old names were carrying. The note about namespace-connection happening in the `Parent` *setter* override rather than the constructor moved up to the module docstring, since it governs every test there.

## Result

```
387 passed, 69 subtests   (361 before: +26 standalone Namespace, 5 relocated net-zero)
```

Verified from a fresh clone of this branch. I also ran the forced-annotation-evaluation sweep over `pyVHDLModel` and the whole test tree — clean, so the renames didn't leave a stale annotation behind on Python 3.11–3.13 (the failure mode from #146).

> [!NOTE]
> Two follow-ups deliberately left out. The `Indexing/` + `DependencyGraph/` split (which requires breaking up `Analyze.py`) is the next step. And `tests/unit/Instantiation/DesignUnit.py` has a pre-existing unused `SelectedName` import — it was already unused before this PR, so I left it rather than mix unrelated cleanup in.


tests/unit/Instantiation/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+#  |_|    |___/                                                                                                        #
+# ==================================================================================================================== #
+# Authors:                                                                                                             #
+#   Patrick Lehmann                                                                                                    #
+#                                                                                                                      #
+# License:                                                                                                             #
+# ==================================================================================================================== #
+# Copyright 2017-2026 Patrick Lehmann - Boetzingen, Germany                                                            #
+# Copyright 2016-2017 Patrick Lehmann - Dresden, Germany                                                               #
```
Paebbels: What is CLAUDE.md saying about the copy right for new files?

tests/unit/Namespace/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+#  |_|    |___/                                                                                                        #
+# ==================================================================================================================== #
+# Authors:                                                                                                             #
+#   Patrick Lehmann                                                                                                    #
+#                                                                                                                      #
+# License:                                                                                                             #
+# ==================================================================================================================== #
+# Copyright 2017-2026 Patrick Lehmann - Boetzingen, Germany                                                            #
+# Copyright 2016-2017 Patrick Lehmann - Dresden, Germany                                                               #
```
Paebbels: here too.

tests/unit/Namespace/DesignUnits.py
```diff
@@ -0,0 +1,57 @@
...
+	print("ERROR: you called a testcase declaration file as an executable module.")
+	print("Use: 'python -m unitest <testcase module>'")
+	exit(1)
+
+
+def _entitySymbol(name: str = "e") -> EntitySymbol:
+	return EntitySymbol(SimpleName(name))
+
+
+class Architectures(TestCase):
```
Paebbels, line 50: We can declare on object of same name at:
* entity port/generic
* entity declartion region
* architecture declaration region
* in generate or block
* in a process
* in a nested subprogram
* in a sequential block statement (VHDL-2019) -> add to a `VHDL-2019.TODOs.md` list

Don't we need to check if hiding elements from an outer scope and nesting scopes works correctly?

tests/unit/Instantiation/__init__.py
Paebbels: CLAUDE.md says new Python files get `Copyright 2026-2026 Patrick Lehmann - Boetzingen, Germany`. Fixed.

The cause: I created both package `__init__.py` files by copying `tests/unit/__init__.py`, so they inherited its `Copyright 2017-2026` line plus the `2016-2017 ... Dresden` line. The three new test *modules* I wrote from scratch (`Namespace.py`, `Statements.py`, `DesignUnits.py`) already carried the correct `2026-2026` header — only the two copied inits were wrong.

tests/unit/Namespace/__init__.py
Paebbels: Fixed the same way — `tests/unit/Namespace/__init__.py` now carries only `Copyright 2026-2026 Patrick Lehmann - Boetzingen, Germany`.

tests/unit/Namespace/DesignUnits.py
Paebbels: Yes — and testing it turned out to be the useful part, because it showed that **only 3 of your 7 scopes can be tested today**. I checked each one empirically rather than assuming.

| Scope | Has a `Namespace`? |
|---|---|
| entity port / generic | ❌ never indexed |
| entity declaration region | ✅ |
| architecture declaration region | ✅ |
| in a generate or block | ✅ |
| in a process | ❌ no namespace at all |
| in a nested subprogram | ❌ no namespace at all |
| sequential block statement (VHDL-2019) | ❌ not modelled |

**New `tests/unit/Namespace/Hiding.py`** covers the three that work — entity → architecture, architecture → block, three levels of nested blocks, and architecture → for-generate. For each: an inner declaration hides a same-named outer one, items declared only outside stay reachable through the whole chain, and the outer scope keeps resolving its own. **Hiding works correctly in all of them** — 12 tests, no bugs found.

One thing worth knowing for anyone writing more of these: the entity → architecture link is made by `Library.LinkArchitectures`, *not* by assigning `Architecture.Parent` (an architecture's parent is its document; the entity relation goes through the entity symbol). My first probe wired it via `.Parent`, saw the namespaces unlinked, and I nearly reported that as a bug — it needs a real linked design. The test builds one.

### Two gaps this exposed

Recorded in `pyVHDLModel.Findings.md` rather than fixed here, since both need a design decision:

1. **Processes and subprograms have no `Namespace`.** `ProcessStatement` and `Procedure`/`Function` aren't `ConcurrentDeclarationRegionMixin`, so a variable declared in a process or nested subprogram can't be resolved at all. All three missing cases — process, subprogram, and the VHDL-2019 sequential block statement — are *sequential* declarative regions, which suggests one shared fix (a sequential counterpart to the concurrent mixin) rather than three separate ones.

2. **Entity ports/generics are never indexed into the namespace.** `IndexDeclaredItems` walks `_declaredItems` only; ports and generics live in `_portItems`/`_genericItems` and never reach `_namespace`. Verified: for `port (clk : in natural)`, after `Entity.IndexDeclaredItems()` the namespace is empty and `FindObject('clk')` raises `ExtendedKeyError`. That's a notable one, since a port name is visible throughout the entity and all its architectures.

### VHDL-2019 list

Created **`VHDL-2019.TODOs.md`** at the repo root with the sequential block statement, cross-referenced to the sequential-regions finding above. I also added `range_expression` — the third alternative of `range ::= range_attribute_name | simple_range | range_expression`, which #145 deliberately left out because no translation path produces one.

Tell me if you'd rather that file lived somewhere else, or be scoped per-repo — VHDL-2019 support spans pyVHDLModel and pyGHDL.dom, and this one only records model-side gaps.

399 passed, verified from a fresh clone of the pushed branch.

Paebbels: Removed — `VHDL-2019.TODOs.md` is out of the PR. Amended the last commit (`090fb65` → `36570b11`) and force-pushed; the PR's file list no longer contains it, and a fresh clone of the branch doesn't either.

The notes still exist, just privately: the file now lives at `/work/VHDL-2019.TODOs.md` alongside `Handoff.md` and the two `Findings.md` notebooks, none of which are in a git repo.

I also removed a reference the committed code was making to those private files. `Hiding.py`'s module docstring said the gaps were "recorded in `pyVHDLModel.Findings.md` and `VHDL-2019.TODOs.md`" — pointing published tests at our private notebooks. It now just states inline which declarative regions do and don't own a namespace, which is the part a reader of the test actually needs.

One caveat worth stating plainly: force-pushing removes the file from the branch, but the old commit `090fb65` may still be retrievable on GitHub by its SHA until it's garbage-collected, and the PR timeline records the force-push. Nothing in the file was sensitive, so I assume that's fine — say the word if you'd rather I took stronger measures.

399 passed, re-verified from a fresh clone.

PR #148, opened 2026-07-26
Namespaces for sequential declaration regions, and for generics/ports/parameters

Paebbels:
Closes the first of the two gaps found in #147's review: processes and subprograms are declarative regions, but neither owned a `Namespace`. A variable declared in a process or a nested subprogram couldn't be resolved at all, and couldn't hide a same-named signal from the scope around it.

## One mixin for both, per the grammar

I checked the [EBNF](https://www.sigasi.com/tech/vhdl2017.ebnf/) rather than assuming: **`process_declarative_item` and `subprogram_declarative_item` are byte-identical**, so one mixin genuinely serves both.

Against `block_declarative_item` (the concurrent region), a sequential region **gains** `variable_declaration` and **loses** `signal_declaration`, `shared_variable_declaration`, `component_declaration`, `mode_view_declaration` and all the specifications. So `SequentialDeclarationRegionMixin` indexes types, subtypes, constants, **variables**, files, functions and procedures — no signals, shared variables or components.

That variable support is the point: a variable is the usual thing you'd hide an outer signal with, and the concurrent mixin explicitly can't index one (it raises a `NotImplementedWarning` there).

## What changed

- **`SequentialDeclarationRegionMixin`** replaces `SequentialDeclarationsMixin`, which held only `_declaredItems` and no namespace. Moved into `Regions.py` so both declaration regions sit side by side, and it now owns `_namespace` itself rather than assuming the host provides one.
- **`Subprogram`** uses it too, resolving its long-standing `# TODO: use mixin class`.
- Both hosts connect their namespace to the enclosing region in the **`Parent` setter**, matching how blocks and generates already do it. `Subprogram` guards with an `isinstance` check, because a subprogram can also be a protected type's method — and `ProtectedType`/`ProtectedTypeBody` are *not* declaration regions, so unguarded chaining would crash there.

## The include-order problem, and how it's resolved

`Subprogram` now imports `Regions` for its base class, but `Regions` already imported `Subprogram` for `Function`/`Procedure` — a genuine cycle, and a base class can't be a lazy import.

Broken by making the `Regions → Subprogram` edge non-module-level: `Function`/`Procedure` are **quoted** in annotations and imported lazily inside `IndexDeclaredItems`. Both halves follow existing in-repo precedent — `Type.py` already writes `List[Union['Procedure', 'Function']]`, and `IndexDeclaredItems` already had a lazy `from pyVHDLModel.DesignUnit import Component`.

Quoted forward references are never evaluated eagerly, so this is safe on 3.11–3.13 too, where class-level annotations *are* evaluated at class-creation time. Confirmed with the forced-annotation-evaluation sweep (clean), which is what caught the analogous problem in #146.

## Testing

New `tests/unit/Namespace/SequentialRegions.py` — 16 tests covering namespace naming (label / normalized identifier / `None` for an unlabelled process), indexing of every declarable kind, nesting into an architecture and into a block, a **process variable hiding an architecture signal**, a **subprogram variable hiding an architecture signal**, **subprograms nesting inside subprograms**, and the searched-namespace chain on a failed lookup.

```
415 passed, 69 subtests   (399 before)
```

Verified from a fresh clone. Also **cross-checked against pyGHDL.dom on `paebbels/dom-updates`: 185 passed, 17 xfailed, 0 failed** — the constructor and `Parent`-setter changes don't regress the translation layer. `SequentialDeclarationsMixin` had no users outside this repo (checked pyGHDL too), so removing it breaks nothing there.

> [!NOTE]
> The second gap from #147's review — entity ports/generics never being indexed into the entity's namespace — is untouched here and still open. It's independent of this change.

### DIFF
--- a/pyVHDLModel/Base.py
+++ b/pyVHDLModel/Base.py
@@ -37,6 +37,7 @@ Base-classes for the VHDL language model.
 from enum                  import unique, Enum
 from typing                import Type, Tuple, List, Iterable, Optional as Nullable, Union, cast
 
+from pyTooling.Common      import getFullyQualifiedName
 from pyTooling.Decorators  import export, readonly
 from pyTooling.MetaClasses import ExtendedType
 
@@ -305,7 +306,9 @@ def identifiersOf(item) -> Tuple[str, ...]:
 	elif isinstance(item, NamedEntityMixin):
 		return (item._identifier, )
 
-	raise TypeError(f"Item '{item!r}' is neither a NamedEntityMixin nor a MultipleNamedEntityMixin.")
+	ex = TypeError(f"Item '{item}' is neither a NamedEntityMixin nor a MultipleNamedEntityMixin.")
+	ex.add_note(f"Got type '{getFullyQualifiedName(item)}'.")
+	raise ex
 
 
 @export
@@ -330,7 +333,9 @@ def normalizedIdentifiersOf(item) -> Tuple[str, ...]:
 	elif isinstance(item, NamedEntityMixin):
 		return (item._normalizedIdentifier, )
 
-	raise TypeError(f"Item '{item!r}' is neither a NamedEntityMixin nor a MultipleNamedEntityMixin.")
+	ex = TypeError(f"Item '{item}' is neither a NamedEntityMixin nor a MultipleNamedEntityMixin.")
+	ex.add_note(f"Got type '{getFullyQualifiedName(item)}'.")
+	raise ex
 
 
 @export
--- a/pyVHDLModel/Base.py
+++ b/pyVHDLModel/Base.py
@@ -282,6 +282,58 @@ class MultipleNamedEntityMixin(metaclass=ExtendedType, mixin=True):
 
 
 @export
+def identifiersOf(item) -> Tuple[str, ...]:
+	"""
+	Return an item's identifier(s), regardless of how many names its declaration carries.
+
+	VHDL entities come in two shapes: singularly named ones deriving from :class:`NamedEntityMixin`
+	(``generic (type T)``, ``GenericProcedureInterfaceItem``, ...) and plurally named ones deriving from
+	:class:`MultipleNamedEntityMixin`, where one declaration names several items at once
+	(``port (p1, p2 : in bit)``, and every ``Constant``/``Signal``/``Variable``/``File``-derived item).
+
+	:param item:      A singularly or plurally named entity.
+	:returns:         The item's identifiers.
+	:raises TypeError: If the item is neither singularly nor plurally named.
+
+	.. seealso::
+
+	   :func:`normalizedIdentifiersOf`
+	     The same, but normalized (lower case) - use that for dictionary keys and name resolution.
+	"""
+	if isinstance(item, MultipleNamedEntityMixin):
+		return item._identifiers
+	elif isinstance(item, NamedEntityMixin):
+		return (item._identifier, )
+
+	raise TypeError(f"Item '{item!r}' is neither a NamedEntityMixin nor a MultipleNamedEntityMixin.")
+
+
+@export
+def normalizedIdentifiersOf(item) -> Tuple[str, ...]:
+	"""
+	Return an item's normalized (lower case) identifier(s).
+
+	This is the form used as dictionary keys and for name resolution, because VHDL identifiers are
+	case-insensitive.
+
+	:param item:      A singularly or plurally named entity.
+	:returns:         The item's normalized identifiers.
+	:raises TypeError: If the item is neither singularly nor plurally named.
+
+	.. seealso::
+
+	   :func:`identifiersOf`
+	     The same, but as written in the source - use that for rendering.
+	"""
+	if isinstance(item, MultipleNamedEntityMixin):
+		return item._normalizedIdentifiers
+	elif isinstance(item, NamedEntityMixin):
+		return (item._normalizedIdentifier, )
+
+	raise TypeError(f"Item '{item!r}' is neither a NamedEntityMixin nor a MultipleNamedEntityMixin.")
+
+
+@export
 class LabeledEntityMixin(metaclass=ExtendedType, mixin=True):
 	"""
 	A ``LabeledEntityMixin`` is a mixin class for all VHDL entities that can have labels.
--- a/pyVHDLModel/Regions.py
+++ b/pyVHDLModel/Regions.py
@@ -41,9 +41,12 @@ from pyTooling.MetaClasses  import ExtendedType
 from pyTooling.Warning      import WarningCollector
 
 from pyVHDLModel.Exception  import NotImplementedWarning
+from pyVHDLModel.Namespace  import Namespace
 from pyVHDLModel.Object     import Constant, SharedVariable, File, Variable, Signal
-from pyVHDLModel.Subprogram import Subprogram, Function, Procedure
 from pyVHDLModel.Type       import Subtype, FullType
+
+# `pyVHDLModel.Subprogram` imports this module (Subprogram is a sequential declaration region), so
+# `Function`/`Procedure` are quoted in annotations and imported lazily where they're needed at runtime.
 
 
 @export
@@ -62,8 +65,8 @@ class ConcurrentDeclarationRegionMixin(metaclass=ExtendedType, mixin=True):
 	_files:           Dict[str, File]                   #: Dictionary of all files declared in this concurrent declaration region.
 	# _subprograms:     Dict[str, List[Subprogram]]  #: Dictionary of all subprograms declared in this concurrent declaration region.
 	# FIXME: overloads are only collected into a list, not matched/resolved by signature.
-	_functions:       Dict[str, List[Function]]         #: Dictionary of all functions declared in this concurrent declaration region, keyed by name; each entry is a list of overloads.
-	_procedures:      Dict[str, List[Procedure]]        #: Dictionary of all procedures declared in this concurrent declaration region, keyed by name; each entry is a list of overloads.
+	_functions:       Dict[str, List['Function']]         #: Dictionary of all functions declared in this concurrent declaration region, keyed by name; each entry is a list of overloads.
+	_procedures:      Dict[str, List['Procedure']]        #: Dictionary of all procedures declared in this concurrent declaration region, keyed by name; each entry is a list of overloads.
 	_components:      Dict[str, Any]                    #: Dictionary of all components declared in this concurrent declaration region.
 
 	def __init__(self, declaredItems: Nullable[Iterable] = None) -> None:
@@ -123,11 +126,11 @@ class ConcurrentDeclarationRegionMixin(metaclass=ExtendedType, mixin=True):
 	# 	return self._subprograms
 
 	@readonly
-	def Functions(self) -> Dict[str, List[Function]]:
+	def Functions(self) -> Dict[str, List['Function']]:
 		return self._functions
 
 	@readonly
-	def Procedures(self) -> Dict[str, List[Procedure]]:
+	def Procedures(self) -> Dict[str, List['Procedure']]:
 		return self._procedures
 
 	@readonly
@@ -165,6 +168,7 @@ class ConcurrentDeclarationRegionMixin(metaclass=ExtendedType, mixin=True):
 		     Iterate all packages in the library and index declared items.
 		"""
 		from pyVHDLModel.DesignUnit import Component
+		from pyVHDLModel.Subprogram import Function, Procedure
 
 		for item in self._declaredItems:
 			if isinstance(item, FullType):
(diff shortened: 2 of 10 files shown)


Paebbels:
Pushed two more commits, so this PR now closes **both** namespace gaps from #147's review. Retitled accordingly.

## `6e70e39` — generics, ports and parameters get indexed

They were never added to any namespace, so a port name — one of the most common things a name resolves to — couldn't be resolved at all. An entity with `port (clk : in bit)` had a completely empty namespace after `IndexDeclaredItems()`.

**They belong in the same namespace as the declared items, not a separate one.** I put this to the GHDL analyzer rather than reasoning about it:

| Construct | GHDL |
|---|---|
| `port (g : in bit)` beside `generic (g : integer)` | **error**: identifier already used |
| `signal x` beside `port (x : in bit)` | **error** |
| block declaration named like a block port | **error** |
| subprogram variable named like a parameter | **error** |
| subprogram parameter named like its own generic | **error** |

All one declarative region, so one namespace. Covers entities (generics + ports), generic packages (generics), blocks (ports) and subprograms (generics + parameters), and handles both identifier shapes — `generic (type T)` is singularly named, every other interface item is plural.

One implementation note worth flagging: the lists are read **by field name** rather than via `isinstance` against `WithGenericsMixin` and friends. Not every host can inherit those mixins — `pyVHDLModel.Interface` needs `Procedure`/`Function` as real base classes for `GenericProcedureInterfaceItem`/`GenericFunctionInterfaceItem`, so `pyVHDLModel.Subprogram` may **never** import `Interface`, and hand-rolls its own `_genericItems`/`_parameterItems`. `ConcurrentBlockStatement` hand-rolls its port list too (with a `# TODO: extract to mixin`). Reading the canonical field names covers every host and needs no import of `Interface` at all — which also keeps this cycle-free. If you'd rather have `isinstance`, the prerequisite is converting `ConcurrentBlockStatement` to `WithPortsMixin`; `Subprogram` structurally cannot.

## `dad024d` — a test from #147 asserted illegal VHDL

While checking the above I found that `test_ArchitectureDeclarationHidesEntityDeclaration`, which I added in #147, claims an architecture declaration *hides* a same-named entity declaration. It doesn't: an entity and its architecture body are a **single** declarative region (LRM 12.1), so GHDL rejects that as `identifier "s" already used for a declaration`. By contrast a *process* variable shadowing an architecture signal is legal and only warns (`-Whide`).

The model resolves it as shadowing — two nested namespaces — so the test passed and looked right. The nesting itself is correct and is exactly what makes ports visible in the architecture; what's missing is duplicate-declaration detection. The test now documents that gap instead of blessing the behaviour. No production change in that commit.

## Verification

```
428 passed, 69 subtests   (399 on dev)
```

13 new interface-item tests, all confirmed to fail without `6e70e39`. Verified from a fresh clone, annotation sweep clean, and **pyGHDL.dom cross-checked: 185 passed, 0 failed**.

> [!NOTE]
> Two follow-ups recorded rather than done: duplicate-declaration detection (the feature that would have caught the bad test), and making `ProtectedType`/`ProtectedTypeBody` declaration regions — GHDL confirms a protected-type method variable hides a protected-body variable with the same `-Whide` warning, so you were right that they're regions. That one needs `ProtectedTypeBody.__init__`'s `declaredItems`-into-`_methods` conflation resolved first, which I didn't want to change unilaterally.

pyVHDLModel/Regions.py
```diff
@@ -40,11 +40,63 @@
...
 from pyVHDLModel.Object     import Constant, SharedVariable, File, Variable, Signal
-from pyVHDLModel.Subprogram import Subprogram, Function, Procedure
 from pyVHDLModel.Type       import Subtype, FullType
 
+# `pyVHDLModel.Subprogram` imports this module (Subprogram is a sequential declaration region), so
+# `Function`/`Procedure` are quoted in annotations and imported lazily where they're needed at runtime.
+
+
+
+def _NormalizedIdentifiersOf(item) -> Iterable[str]:
```
Paebbels:
```suggestion
def normalizedIdentifiersOf(item) -> Iterable[str]:
```

Shouldn't we place this in a more general module? I think some/many testcases have a similar but duplicated implementation. It could be reused there.

pyVHDLModel/Regions.py
```diff
@@ -40,11 +40,63 @@
...
+	   those mixins: :mod:`pyVHDLModel.Interface` needs :class:`~pyVHDLModel.Subprogram.Procedure` and
+	   :class:`~pyVHDLModel.Subprogram.Function` as real base classes for its generic subprogram interface
+	   items, so :mod:`pyVHDLModel.Subprogram` may never import it and declares its own
+	   ``_genericItems``/``_parameterItems``. Reading the canonical field names covers both cases without
+	   importing :mod:`pyVHDLModel.Interface` at all.
+
+	:param region: The declaration region whose interface items are to be indexed.
+	"""
+	for attributeName in ("_genericItems", "_portItems", "_parameterItems"):
+		for item in getattr(region, attributeName, None) or ():
```
Paebbels: I don't considder using `hasattr` good programming style in my code. The code has a good class hierarchy and can use isinstance or issubclass.

pyVHDLModel/Regions.py
```diff
@@ -40,11 +40,63 @@
...
+	:param item: The declared or interface item.
+	:returns:    The item's normalized identifiers.
+	"""
+	if isinstance(item, MultipleNamedEntityMixin):
+		return item._normalizedIdentifiers
+
+	return (item._normalizedIdentifier, )
+
+
+def _IndexInterfaceItemsInto(region) -> None:
```
Paebbels: I think it's best to have 3 distinct implementations. I try to deduplicate, but I also keep speed in mind. Please apply this algorithm where its needed specific for generics, ports and parameters.

pyVHDLModel/Regions.py
Paebbels: Renamed to `normalizedIdentifiersOf` and moved to `Base.py`, next to the `NamedEntityMixin`/`MultipleNamedEntityMixin` pair it discriminates.

**You were right about the duplicate** — `Interface.py:51` had `_identifiersOf`, doing the same shape discrimination for the as-written identifiers. Removed; its three call sites (`GenericGroup`/`PortGroup`/`ParameterGroup.__str__`) now use a new `identifiersOf` from `Base.py`. So the two variants sit side by side:

- `identifiersOf(item)` — as written, for rendering
- `normalizedIdentifiersOf(item)` — lower case, for dictionary keys and name resolution

Both now raise `TypeError` if the item is neither shape, rather than silently assuming the singular one (the old `_identifiersOf` fell through to `item.Identifier`).

One correction to the expectation, though: I checked `tests/unit` and found **no** duplicated implementation there. The helpers in the test modules (`_signal`, `_port`, `_variable`, `_subtypeSymbol`) are fixture builders, not identifier-shape logic. There is duplication among *those* across test modules, but that is a different cleanup and I have left it alone.

pyVHDLModel/Regions.py
Paebbels: Agreed, and removed — no `getattr`/`hasattr` left in `Regions.py`.

To make `isinstance` actually cover every host, `ConcurrentBlockStatement` now inherits `WithPortsMixin`. It was hand-rolling the identical list and property with a `# TODO: extract to mixin`, so that TODO is resolved too.

One case genuinely cannot use those mixins, so it dispatches on `Subprogram` instead: `pyVHDLModel.Interface` needs `Procedure`/`Function` as **real base classes** for `GenericProcedureInterfaceItem`/`GenericFunctionInterfaceItem`, so `pyVHDLModel.Subprogram` may never import `Interface`. It declares `_genericItems`/`_parameterItems` itself. That is an `isinstance` check against a class in the hierarchy, not attribute probing — but if you would rather have `Subprogram` inherit the mixins, the prerequisite is moving those two interface-item classes out of `Interface.py` to break that edge. Happy to do that if you want it.

pyVHDLModel/Regions.py
Paebbels: Done — three distinct implementations, each applied only where it is needed:

```python
def _IndexGenericItems(region) -> None: ...
def _IndexPortItems(region) -> None: ...
def _IndexParameterItems(region) -> None: ...
```

`ConcurrentDeclarationRegionMixin` calls the generics one for `WithGenericsMixin` hosts and the ports one for `WithPortsMixin` hosts — entities have both, generic packages only generics, blocks only ports. `SequentialDeclarationRegionMixin` calls generics + parameters for `Subprogram`, the only sequential region with interface items. No host pays for a list it does not have, and there is no loop over attribute names any more.

pyVHDLModel/Base.py
```diff
@@ -281,6 +281,58 @@ def NormalizedIdentifiers(self) -> Tuple[str]:
...
+
+	   :func:`normalizedIdentifiersOf`
+	     The same, but normalized (lower case) - use that for dictionary keys and name resolution.
+	"""
+	if isinstance(item, MultipleNamedEntityMixin):
+		return item._identifiers
+	elif isinstance(item, NamedEntityMixin):
+		return (item._identifier, )
+
+	raise TypeError(f"Item '{item!r}' is neither a NamedEntityMixin nor a MultipleNamedEntityMixin.")
```
Paebbels: Please give a hint what type was received see __init__.py in pyVHDLModel at line 2640:
```
		if not isinstance(item, Entity):
			ex = TypeError(f"Parameter 'item' is not of type 'Entity'.")
			if version_info >= (3, 11):  # pragma: no cover
				ex.add_note(f"Got type '{getFullyQualifiedName(item)}'.")
			raise ex
```
It can be used like so:
```
ex = TypeError(f"Item '{item}' is neither a NamedEntityMixin nor a MultipleNamedEntityMixin.")
ex.add_note(f"Got type '{getFullyQualifiedName(item)}'.")
raise ex
```
The Python version guard isn't needed anymore.

At to our findings list to:
* rewrite all raise TypeError with this pattern.
* remove all Python version guards, because the package is 3.11+

Other todos:
* check all read-only properties if marked as `@readonly`
* Add a doc-string to all read-only properties following a pattern we need to discuss.
* Add a doc-string to all other properties.
* Recheck existing properties (incl. readonly) for a outdated/correct doc-string.

pyVHDLModel/Regions.py
```diff
@@ -165,6 +191,16 @@ def IndexDeclaredItems(self) -> None:
...
+		from pyVHDLModel.Interface  import WithGenericsMixin, WithPortsMixin
+		from pyVHDLModel.Subprogram import Function, Procedure
+
+		# An interface item shares the declarative region of the declarative part beside it, so generics and
+		# ports go into this very namespace. Entities have both, generic packages only generics, blocks only
+		# ports.
+		if isinstance(self, WithGenericsMixin):
+			_IndexGenericItems(self)
+		if isinstance(self, WithPortsMixin):
+			_IndexPortItems(self)
```
Paebbels: I would put this into the derived class that knows if generics and ports exist. it can call the helpers and then call super().IndexDeclaredItems()

I f we want to share these helpers, why not derive both regionMixins from the same RegionBaseMixin and have these functions added there as methods (names are ok).

pyVHDLModel/Base.py
Paebbels: Applied to both new functions:

```python
ex = TypeError(f"Item '{item}' is neither a NamedEntityMixin nor a MultipleNamedEntityMixin.")
ex.add_note(f"Got type '{getFullyQualifiedName(item)}'.")
raise ex
```

Verified: message `Item '42' is neither a NamedEntityMixin nor a MultipleNamedEntityMixin.`, note `Got type 'int'.`. No version guard.

All four sweeps are recorded in the findings list, with sizes measured so they can be scheduled:

| Sweep | Size on `dev` + this PR |
|---|---|
| Rewrite `raise TypeError` with the message + note pattern | **14 sites** — `Namespace.py` (8), `__init__.py` (6) |
| Remove Python version guards (package is 3.11+) | **~11 guards**, all in `__init__.py`, plus the `from sys import version_info` |
| Check every read-only property is marked `@readonly` | **109** plain `@property` left to triage |
| Doc-strings on all properties | **244** undocumented, 78 documented |

For the doc-string pattern I have written down exactly what you specified:

- read-only property *exposing a field* (no computation): ``Read-only property to access ... (:attr:`_...`).``
- read-only property that *computes*: ``Read-only property to return ... .``, naming the fields used in the calculation with ``:attr:`...` ``

Useful data point for that sweep: **54 of the 78 existing doc-strings already match the first form**, so the pattern is mostly a matter of filling gaps rather than rewriting. The other 24 ("Returns a list of ...", "The identifier ...") need rechecking against the pattern and for stale content in the same pass. Every property this PR adds is plain field exposure and already uses the "access" form.

pyVHDLModel/Regions.py
Paebbels: Both done.

**`DeclarationRegionMixin`** is now the shared base of `ConcurrentDeclarationRegionMixin` and `SequentialDeclarationRegionMixin`, carrying the three indexers as methods (names kept). It also absorbed `_IndexOtherDeclaredItem`, which both regions had been defining identically.

**The dispatch moved into the derived classes**, so there is no `isinstance` in `IndexDeclaredItems` at all any more — each class knows statically what it has:

| Class | calls |
|---|---|
| `Entity` | `_IndexGenericItems()` + `_IndexPortItems()` |
| `Package` | `_IndexGenericItems()` |
| `ConcurrentBlockStatement` | `_IndexPortItems()` |
| `Subprogram` | `_IndexGenericItems()` + `_IndexParameterItems()` |

each followed by `super().IndexDeclaredItems()`. `Architecture` and `PackageBody` correctly define no override.

This also resolves the wrinkle from the previous round: `Regions.py` no longer imports `Interface` **or** `Subprogram` for this, so the `Subprogram`-cannot-import-`Interface` cycle stops being relevant here — `Subprogram` simply calls the inherited helpers on itself.

428 passed, fresh-clone verified, annotation sweep clean, and pyGHDL.dom cross-checked at 185 passed.

PR #149, opened 2026-07-26
Exception pattern, @readonly marking and property doc-strings

Paebbels:
Three of the four consistency sweeps agreed in #148's review, in three reviewable commits. No behaviour changes beyond the exception notes.

## `29988df` — TypeError pattern + version guards

Every `raise TypeError` now names the offending value in the message and puts detail in a note, matching the reference implementation in `__init__.py`.

- **`Namespace.py` (8 sites)** — the wrong-kind branches report the type actually found. The "was not expected" branches additionally report what the symbol would have accepted, which is the `add_note`-for-expected-items idea that was sitting in the findings list:

  ```
  TypeError: Found signal 'x', but it was not expected.
    Got type 'pyVHDLModel.Object.Signal'.
    Expected one of: PossibleReference.Constant.
  ```

- **`__init__.py` (6 sites)** — the `VHDLVersion` comparison operators report the second operand's type.

The **11 `if version_info >= (3, 11):`** guards around `add_note` are gone, along with the now-unused `from sys import version_info`. Test assertions added for all three note shapes.

## `36c5563` — `@readonly` marking

**76 properties** were a plain `@property` with no setter anywhere, so assignment already failed — just with a less clear error and without the decorator stating the intent.

Selection was conservative: a property is converted only when **no** class anywhere in the package defines a setter of that name, so cases where a subclass supplies the setter are left alone. After this, every remaining plain `@property` (33) genuinely has a setter.

## `42b39e2` — doc-strings on all 244 properties

Following the pattern you specified:

| Kind | Form |
|---|---|
| exposes a field | ``Read-only property to access the ... (:attr:`_...`).`` |
| computes a value | ``Read-only property to return ... .``, naming the fields used |
| settable | ``Property to access/set the ... (:attr:`_...`).`` |

209 are plain `return self._field` and were generated from the property name plus the field they expose. The **six computed** ones (`GenericCount`, `PortCount`, `ParameterCount`, `IsResolved`, `IsPure`, `ParenthesisExpression.Operand`) and the **28 setters** were written individually. I also reviewed the generated text and rewrote `WaveformElement.After`, which came out as "the after".

> [!NOTE]
> Three getters are described as settable rather than read-only because they define a setter **despite being decorated `@readonly`**: `DesignUnit.Document`, `Namespace.ParentNamespace`, `LibraryReferenceSymbol.Library`. I checked on `dev`: the combination works at runtime (assignment succeeds), so this commit only corrects the wording. The contradiction itself is recorded in the findings list — `LibraryReferenceSymbol.Library` is how a library symbol gets *resolved*, so it is meant to be settable and the decorator looks like the mistake.

## Verification

```
428 passed, 69 subtests
```

Fresh clone verified, annotation sweep clean, and pyGHDL.dom cross-checked at **185 passed, 0 failed**.

The fourth sweep from that review — rechecking the 24 pre-existing doc-strings that don't match the pattern ("Returns a list of …", "The identifier …") for wording and staleness — is deliberately **not** in here; it needs reading each one rather than a mechanical pass.

### DIFF
--- a/pyVHDLModel/Common.py
+++ b/pyVHDLModel/Common.py
@@ -87,6 +87,7 @@ class AllowBlackboxMixin(metaclass=ExtendedType, mixin=True):
 
 	@AllowBlackbox.setter
 	def AllowBlackbox(self, value: Nullable[bool]) -> None:
+		"""Property to set the allow blackbox (:attr:`_allowBlackbox`)."""
 		self._allowBlackbox = value
 
 
@@ -118,10 +119,12 @@ class ProcedureCallMixin(metaclass=ExtendedType, mixin=True):
 
 	@readonly
 	def Procedure(self) -> Symbol:
+		"""Read-only property to access the procedure (:attr:`_procedure`)."""
 		return self._procedure
 
 	@readonly
 	def ParameterAssociationItems(self) -> List[ParameterAssociationItem]:
+		"""Read-only property to access the parameter association items (:attr:`_parameterAssociationItems`)."""
 		return self._parameterAssociationItems
 
 
@@ -137,6 +140,7 @@ class AssignmentMixin(metaclass=ExtendedType, mixin=True):
 
 	@readonly
 	def Target(self) -> Symbol:
+		"""Read-only property to access the target (:attr:`_target`)."""
 		return self._target
 
 
@@ -146,6 +150,7 @@ class SignalAssignmentMixin(AssignmentMixin, mixin=True):
 
 	@readonly
 	def Target(self) -> SignalSymbol:
+		"""Read-only property to access the target (:attr:`_target`)."""
 		return self._target
 
 
@@ -164,10 +169,12 @@ class VariableAssignmentMixin(AssignmentMixin, mixin=True):
 
 	@readonly
 	def Target(self) -> VariableSymbol:
+		"""Read-only property to access the target (:attr:`_target`)."""
 		return self._target
 
 	@readonly
 	def Expression(self) -> ExpressionUnion:
+		"""Read-only property to access the expression (:attr:`_expression`)."""
 		return self._expression
 
 
@@ -185,6 +192,7 @@ class WaveformMixin(metaclass=ExtendedType, mixin=True):
 
 	@readonly
 	def Waveform(self) -> List[WaveformElement]:
+		"""Read-only property to access the waveform (:attr:`_waveform`)."""
 		return self._waveform
 
 
@@ -200,6 +208,7 @@ class ExpressionMixin(metaclass=ExtendedType, mixin=True):
 
 	@readonly
 	def Expression(self) -> ExpressionUnion:
+		"""Read-only property to access the expression (:attr:`_expression`)."""
 		return self._expression
 
 
@@ -270,6 +279,7 @@ class ConditionalWaveformsMixin(metaclass=ExtendedType, mixin=True):
 
 	@readonly
 	def ConditionalWaveforms(self) -> List[ConditionalWaveform]:
+		"""Read-only property to access the conditional waveforms (:attr:`_conditionalWaveforms`)."""
 		return self._conditionalWaveforms
 
 
@@ -357,6 +367,7 @@ class SelectedWaveformsMixin(metaclass=ExtendedType, mixin=True):
 
 	@readonly
 	def SelectedWaveforms(self) -> List[Union[SelectedWaveform, OthersSelectedWaveform]]:
+		"""Read-only property to access the selected waveforms (:attr:`_selectedWaveforms`)."""
 		return self._selectedWaveforms
 
 
@@ -377,4 +388,5 @@ class SelectedExpressionsMixin(metaclass=ExtendedType, mixin=True):
 
 	@readonly
 	def SelectedExpressions(self) -> List[Union[SelectedExpression, OthersSelectedExpression]]:
+		"""Read-only property to access the selected expressions (:attr:`_selectedExpressions`)."""
 		return self._selectedExpressions
--- a/pyVHDLModel/Common.py
+++ b/pyVHDLModel/Common.py
@@ -120,7 +120,7 @@ class ProcedureCallMixin(metaclass=ExtendedType, mixin=True):
 	def Procedure(self) -> Symbol:
 		return self._procedure
 
-	@property
+	@readonly
 	def ParameterAssociationItems(self) -> List[ParameterAssociationItem]:
 		return self._parameterAssociationItems
 
--- a/pyVHDLModel/Common.py
+++ b/pyVHDLModel/Common.py
@@ -65,7 +65,8 @@ class AllowBlackboxMixin(metaclass=ExtendedType, mixin=True):
 	@property
 	def AllowBlackbox(self) -> bool:
 		"""
-		Read-only property to check if a design supports blackboxes (:attr:`_allowBlackbox`).
+		Property to return whether a design supports blackboxes, inherited from the parent if not set locally
+		(:attr:`_allowBlackbox`).
 
 		.. rubric:: Algorithm
 
@@ -87,7 +88,6 @@ class AllowBlackboxMixin(metaclass=ExtendedType, mixin=True):
 
 	@AllowBlackbox.setter
 	def AllowBlackbox(self, value: Nullable[bool]) -> None:
-		"""Property to set the allow blackbox (:attr:`_allowBlackbox`)."""
 		self._allowBlackbox = value
 
 
--- a/pyVHDLModel/Concurrent.py
+++ b/pyVHDLModel/Concurrent.py
@@ -192,9 +192,9 @@ class ComponentInstantiation(Instantiation):
 		self._component = componentSymbol
 		componentSymbol.Parent = self
 
-	@property
+	@readonly
 	def Component(self) -> ComponentInstantiationSymbol:
-		"""Property to access the component (:attr:`_component`)."""
+		"""Read-only property to access the component (:attr:`_component`)."""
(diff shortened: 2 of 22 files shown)


pyVHDLModel/Common.py
```diff
@@ -87,6 +87,7 @@ def AllowBlackbox(self) -> bool:
 
 	@AllowBlackbox.setter
 	def AllowBlackbox(self, value: Nullable[bool]) -> None:
+		"""Property to set the allow blackbox (:attr:`_allowBlackbox`)."""
```
Paebbels: Do we need to document setters? Are these visible in Sphinx. I don't think so.

pyVHDLModel/Concurrent.py
```diff
@@ -191,6 +194,7 @@ def __init__(
 
 	@property
```
Paebbels: Isn't this read-only?

pyVHDLModel/Concurrent.py
```diff
@@ -265,6 +271,7 @@ def __init__(
...
```
Paebbels: is it read-only?

pyVHDLModel/DesignUnit.py
```diff
@@ -541,10 +555,12 @@ def __init__(
 
 	@property
```
Paebbels: isn't this read-only?

pyVHDLModel/DesignUnit.py
```diff
@@ -664,6 +681,7 @@ def __init__(
...
```
Paebbels: this is also read-only, right?

pyVHDLModel/Exception.py
```diff
@@ -298,10 +309,12 @@ def __init__(self, configuration: 'Configuration', library: 'Library') -> None:
 
 	@property
```
Paebbels: read-only?

pyVHDLModel/Exception.py
```diff
@@ -298,10 +309,12 @@ def __init__(self, configuration: 'Configuration', library: 'Library') -> None:
...
 	def Library(self) -> 'Library':
+		"""Property to access the library (:attr:`_library`)."""
 		return self._library
 
 	@property
```
Paebbels: read-only?

pyVHDLModel/Exception.py
```diff
@@ -329,10 +342,12 @@ def __init__(self, context: 'Context', library: 'Library') -> None:
...
```
Paebbels: read-only?

pyVHDLModel/Exception.py
```diff
@@ -329,10 +342,12 @@ def __init__(self, context: 'Context', library: 'Library') -> None:
...
 	def Library(self) -> 'Library':
+		"""Property to access the library (:attr:`_library`)."""
 		return self._library
 
 	@property
```
Paebbels: read-only?

pyVHDLModel/Exception.py
```diff
@@ -358,10 +373,12 @@ def __init__(self, context: 'Context', librarySymbol: Symbol) -> None:
 		self._librarySymbol = librarySymbol
 		self._context = context
 
-	@property
+	@readonly
 	def LibrarySymbol(self) -> Symbol:
+		"""Read-only property to access the library symbol (:attr:`_librarySymbol`)."""
 		return self._librarySymbol
 
 	@property
```
Paebbels: read-only?

pyVHDLModel/Expression.py
```diff
@@ -782,6 +806,7 @@ def __init__(self, subtype: Symbol, parent: Nullable[ModelEntity] = None) -> Non
 
 	@property
```
Paebbels: read-only?

pyVHDLModel/Symbol.py
```diff
@@ -156,10 +159,12 @@ def __init__(self, name: Name) -> None:
 
 	@readonly
```
Paebbels: this is read/write, right?

pyVHDLModel/Common.py
Paebbels: You are right, and I checked before removing rather than assuming — `property.__doc__` is the **getter's** doc-string; the setter's is only reachable as `.fset.__doc__`, which neither Sphinx autodoc nor pydoc renders:

```python
C.P.__doc__       -> "GETTER docstring."
C.P.fset.__doc__  -> "SETTER docstring."
pydoc shows SETTER: False
```

All 28 setter doc-strings removed.

pyVHDLModel/Symbol.py
Paebbels: Yes — read/write. It is `@readonly` **and** defines a setter, which is how a library symbol gets *resolved*, so settable is clearly the intent and the decorator was the mistake. Changed to `@property`.

Two more had the same contradiction and are also `@property` now: `DesignUnit.Document` and `Namespace.ParentNamespace`.

pyVHDLModel/Concurrent.py
Paebbels: Yes, read-only — and so were all the others you flagged. Thanks for catching this; **my selection rule was wrong**, not just conservative.

The first sweep skipped a property whenever *any* class in the package defined a setter of that name. `ComponentInstantiation.Component` was therefore left settable purely because `ComponentInstantiationSymbol` happens to have a `Component` setter — a completely unrelated class.

Now done **per class**: a getter is `@readonly` unless its own class defines the setter. That converts 13 more, covering every line you flagged plus two you did not (`EntityInstantiation.Architecture`, `RecordTypeElement.Subtype`), which fall out of the same rule:

| File | Properties |
|---|---|
| `Concurrent.py` | `Component`, `Entity`, `Architecture`, `Configuration` |
| `DesignUnit.py` | `PackageBody.Package`, `Architecture.Entity` |
| `Exception.py` | `Library` ×2, `Configuration`, `Context` ×2 |
| `Expression.py` | `SubtypeAllocation.Subtype` |
| `Type.py` | `RecordTypeElement.Subtype` |

I re-analysed the whole package afterwards to confirm consistency: no `@property` lacking a setter in its own class, no `@readonly` with one, and no doc-string contradicting its decorator. That last check also caught `AllowBlackbox`, whose pre-existing doc-string claimed read-only while having a setter — reworded.

428 passed, pyGHDL.dom cross-checked at 185 passed.

pyVHDLModel/Concurrent.py
Paebbels: Yes — fixed. See the [explanation on the first of these](https://github.com/VHDL/pyVHDLModel/pull/149#discussion_r3653448430): my selection rule matched setter names across the whole package instead of per class, so these were wrongly left settable. All are `@readonly` now.

pyVHDLModel/DesignUnit.py
Paebbels: Yes — fixed. See the [explanation on the first of these](https://github.com/VHDL/pyVHDLModel/pull/149#discussion_r3653448430): my selection rule matched setter names across the whole package instead of per class, so these were wrongly left settable. All are `@readonly` now.

pyVHDLModel/DesignUnit.py
Paebbels: Yes — fixed. See the [explanation on the first of these](https://github.com/VHDL/pyVHDLModel/pull/149#discussion_r3653448430): my selection rule matched setter names across the whole package instead of per class, so these were wrongly left settable. All are `@readonly` now.

pyVHDLModel/Exception.py
Paebbels: Yes — fixed. See the [explanation on the first of these](https://github.com/VHDL/pyVHDLModel/pull/149#discussion_r3653448430): my selection rule matched setter names across the whole package instead of per class, so these were wrongly left settable. All are `@readonly` now.

pyVHDLModel/Exception.py
Paebbels: Yes — fixed. See the [explanation on the first of these](https://github.com/VHDL/pyVHDLModel/pull/149#discussion_r3653448430): my selection rule matched setter names across the whole package instead of per class, so these were wrongly left settable. All are `@readonly` now.

pyVHDLModel/Exception.py
Paebbels: Yes — fixed. See the [explanation on the first of these](https://github.com/VHDL/pyVHDLModel/pull/149#discussion_r3653448430): my selection rule matched setter names across the whole package instead of per class, so these were wrongly left settable. All are `@readonly` now.

pyVHDLModel/Exception.py
Paebbels: Yes — fixed. See the [explanation on the first of these](https://github.com/VHDL/pyVHDLModel/pull/149#discussion_r3653448430): my selection rule matched setter names across the whole package instead of per class, so these were wrongly left settable. All are `@readonly` now.

pyVHDLModel/Exception.py
Paebbels: Yes — fixed. See the [explanation on the first of these](https://github.com/VHDL/pyVHDLModel/pull/149#discussion_r3653448430): my selection rule matched setter names across the whole package instead of per class, so these were wrongly left settable. All are `@readonly` now.

pyVHDLModel/Expression.py
Paebbels: Yes — fixed. See the [explanation on the first of these](https://github.com/VHDL/pyVHDLModel/pull/149#discussion_r3653448430): my selection rule matched setter names across the whole package instead of per class, so these were wrongly left settable. All are `@readonly` now.

PR #150, opened 2026-07-26
Bring the remaining property doc-strings onto the pattern, and fix stale ones

Paebbels:
Completes the property doc-string work: the 24 doc-strings that predated #149 and didn't match the agreed pattern. Doc-strings only — no code changes.

## Rewritten to the pattern

- **Field accessors** → ``Read-only property to access ... (:attr:`_...`).`` — the nine identifier/label/documentation mixins in `Base.py`, four in `Name.py`, `ScalarConstraint.Constraint`, `Component.IsBlackbox`, and the `Library.*` collections.
- **Computed** → ``Read-only property to return ... .``, naming the field used — `Name.HasPrefix` (:attr:`_prefix` is set), `VHDLVersion.IsVHDL`, `VHDLVersion.IsAMS`.
- `OptionallyNamedEntityMixin` now says its identifier may be `None`, which is the whole difference from `NamedEntityMixin` and wasn't mentioned before (both carried identical text).

## Rechecking the content found two real errors

This is why the pass was worth doing separately rather than mechanically:

**1. A dangling cross-reference.** `Component.IsBlackbox` documented ``:attr:`_isBlackbox``` — but the field is `_isBlackBox`. Sphinx would not have resolved it.

**2. Fifteen properties described a `Dict` as a "list".** `Library` and `Document` each expose `Contexts`/`Configurations`/`Entities`/`Architectures`/`Packages`/`PackageBodies` (plus `Document`'s three `Verification*` ones) as `Dict[str, ...]`, while both the summary line and `:returns:` said *list*:

```python
def Contexts(self) -> Dict[str, Context]:
    """Returns a list of all context declarations declared in this library."""
```

They now say dictionary and state the key. `Architectures` is `Dict[str, Dict[str, Architecture]]`, so its `:returns:` spells out both key levels — that one was doubly misleading.

> [!NOTE]
> Only **6 of those 15** were in the off-pattern set. The other 9 already matched the pattern and were wrong only in content, so a pattern-based sweep would have missed them entirely. Worth remembering if similar sweeps come up: matching the template and being correct are independent properties.

## Verification

Re-analysed the whole package afterwards:

- properties still off-pattern: **0**
- properties describing a `Dict` as a list: **0**
- dangling ``:attr:`` references: **0**

428 passed, verified from a fresh clone, annotation sweep clean, pyGHDL.dom cross-checked at **185 passed, 0 failed**.

With this, all four consistency sweeps agreed in #148's review are complete.

### DIFF
--- a/pyVHDLModel/DesignUnit.py
+++ b/pyVHDLModel/DesignUnit.py
@@ -524,24 +524,24 @@ class Package(PrimaryUnit, DesignUnitWithContextMixin, WithGenericsMixin, Concur
 		return self._declaredItems
 
 	@readonly
-	def DeferredConstants(self):
+	def DeferredConstants(self) -> Dict[str, DeferredConstant]:
 		"""
 		Read-only property to access the deferred constants (:attr:`_deferredConstants`).
 
-		:returns: The deferred constants.
+		:returns: Dictionary of deferred constants, indexed by normalized identifier.
 		"""
 		return self._deferredConstants
 
 	@readonly
-	def Components(self):
+	def Components(self) -> Dict[str, 'Component']:
 		"""
 		Read-only property to access the components (:attr:`_components`).
 
-		:returns: The components.
+		:returns: Dictionary of components, indexed by normalized identifier.
 		"""
 		return self._components
 
-	def _IndexOtherDeclaredItem(self, item):
+	def _IndexOtherDeclaredItem(self, item) -> None:
 		if isinstance(item, DeferredConstant):
 			for normalizedIdentifier in item.NormalizedIdentifiers:
 				self._deferredConstants[normalizedIdentifier] = item
--- a/pyVHDLModel/DesignUnit.py
+++ b/pyVHDLModel/DesignUnit.py
@@ -750,9 +750,12 @@ class Component(ModelEntity, NamedEntityMixin, DocumentedEntityMixin, AllowBlack
 	@readonly
 	def IsBlackbox(self) -> Nullable[bool]:
 		"""
-		Read-only property to access whether the component is a blackbox (:attr:`_isBlackBox`).
+		Check if the component is a blackbox (:attr:`_isBlackBox`).
 
-		:returns: ``True``, if the component is a blackbox.
+		If components were not linked to matching entities, this property returns ``None``.
+
+		:returns: ``True``, if the component is a blackbox; ``False``, if it is not; ``None``, if components
+		          were not linked to entities yet.
 		"""
 		return self._isBlackBox
 
--- a/pyVHDLModel/DesignUnit.py
+++ b/pyVHDLModel/DesignUnit.py
@@ -229,7 +229,11 @@ class DesignUnit(ModelEntity, NamedEntityMixin, DocumentedEntityMixin):
 
 	@property
 	def Document(self) -> 'Document':
-		"""Property to access the document (:attr:`_document`)."""
+		"""
+		Property to access the document (:attr:`_document`).
+
+		:returns: The document.
+		"""
 		return self._document
 
 	@Document.setter
@@ -238,7 +242,11 @@ class DesignUnit(ModelEntity, NamedEntityMixin, DocumentedEntityMixin):
 
 	@property
 	def Library(self) -> 'Library':
-		"""Property to access the library (:attr:`_parent`)."""
+		"""
+		Property to access the library (:attr:`_parent`).
+
+		:returns: The library.
+		"""
 		return self._parent
 
 	@Library.setter
@@ -284,17 +292,29 @@ class DesignUnit(ModelEntity, NamedEntityMixin, DocumentedEntityMixin):
 
 	@readonly
 	def ReferencedLibraries(self) -> Dict[str, 'Library']:
-		"""Read-only property to access the referenced libraries (:attr:`_referencedLibraries`)."""
+		"""
+		Read-only property to access the referenced libraries (:attr:`_referencedLibraries`).
+
+		:returns: Dictionary of referenced libraries, indexed by normalized identifier.
+		"""
 		return self._referencedLibraries
 
 	@readonly
 	def ReferencedPackages(self) -> Dict[str, 'Package']:
-		"""Read-only property to access the referenced packages (:attr:`_referencedPackages`)."""
+		"""
+		Read-only property to access the referenced packages (:attr:`_referencedPackages`).
+
+		:returns: Dictionary of referenced packages, indexed by normalized identifier.
+		"""
 		return self._referencedPackages
 
 	@readonly
 	def ReferencedContexts(self) -> Dict[str, 'Context']:
-		"""Read-only property to access the referenced contexts (:attr:`_referencedContexts`)."""
+		"""
+		Read-only property to access the referenced contexts (:attr:`_referencedContexts`).
+
+		:returns: Dictionary of referenced contexts, indexed by normalized identifier.
+		"""
 		return self._referencedContexts
 
 	@readonly
@@ -403,17 +423,29 @@ class Context(PrimaryUnit):
 
 	@readonly
 	def LibraryReferences(self) -> List[LibraryClause]:
-		"""Read-only property to access the library references (:attr:`_libraryReferences`)."""
+		"""
+		Read-only property to access the library references (:attr:`_libraryReferences`).
+
+		:returns: List of library references.
+		"""
 		return self._libraryReferences
 
 	@readonly
 	def PackageReferences(self) -> List[UseClause]:
-		"""Read-only property to access the package references (:attr:`_packageReferences`)."""
+		"""
+		Read-only property to access the package references (:attr:`_packageReferences`).
+
+		:returns: List of package references.
+		"""
 		return self._packageReferences
 
 	@readonly
 	def ContextReferences(self) -> List[ContextReference]:
-		"""Read-only property to access the context references (:attr:`_contextReferences`)."""
+		"""
+		Read-only property to access the context references (:attr:`_contextReferences`).
+
+		:returns: List of context references.
+		"""
 		return self._contextReferences
 
 	def __str__(self) -> str:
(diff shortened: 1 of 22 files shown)


pyVHDLModel/__init__.py
```diff
@@ -283,18 +283,18 @@ def __hash__(self) -> int:
 	@readonly
 	def IsVHDL(self) -> bool:
 		"""
-		Checks if the version is a VHDL (not VHDL-AMS) version.
+		Read-only property to return whether the version is a VHDL (not VHDL-AMS) version.
```
Paebbels: For read-only properties starting with `Is***`, we should use the `Check if ...` sentence.

Add to the handoff documentation.

pyVHDLModel/__init__.py
```diff
@@ -283,18 +283,18 @@ def __hash__(self) -> int:
...
 
-		:returns:          True if version is a VHDL version.
+		:returns: ``True``, if the version is a VHDL version.
```
Paebbels, line 288: Check for all boolean returning functions and methods if the the :returns: documentation starts with `\`\`True\`\`, if ...` or not so likely `\`\`False\`\`, if ...`.
Add to the handoff documentation.

pyVHDLModel/__init__.py
```diff
@@ -2921,81 +2946,82 @@ def DesignUnits(self) -> List[DesignUnit]:
 	@readonly
 	def Contexts(self) -> Dict[str, Context]:
 		"""
-		Read-only property to access a list of all context declarations found in this document (:attr:`_contexts`).
+		Read-only property to access the dictionary of all context declarations in this document (:attr:`_contexts`).
 
-		:returns: List of all contexts.
+		:returns: Dictionary of all contexts, keyed by normalized identifier.
```
Paebbels: is `keyed by` good english? I suggest `indexed by`.

pyVHDLModel/DesignUnit.py
```diff
@@ -750,11 +750,9 @@ def __init__(
 	@readonly
 	def IsBlackbox(self) -> Nullable[bool]:
 		"""
-		Read-only property returning true, if this component is a blackbox (:attr:`_isBlackbox`).
+		Read-only property to access whether the component is a blackbox (:attr:`_isBlackBox`).
 
-		If components were not linked to matching entities, this property returns None.
```
Paebbels, line 755: you removed the essential explanation when it returns None. The function has 3 possible return values: true, false and none. None must be documented too.

pyVHDLModel/Name.py
```diff
@@ -77,7 +77,7 @@ def Identifier(self) -> str:
 	@readonly
 	def NormalizedIdentifier(self) -> str:
 		"""
-		The normalized identifier the name is referencing.
+		Read-only property to access the normalized identifier this name references (:attr:`_normalizedIdentifier`).
```
Paebbels, line 81: Should we have a seealso entry for "normalized" linking to the glossary so we can explain normalized identifiers?

Can be added to the findings file.

pyVHDLModel/__init__.py
Paebbels: Done, and written into the handoff.

Six properties now use the `Check if ...` sentence instead of the generic lead: `IsVHDL`, `IsAMS`, `IsPure`, `IsResolved`, `IsBlackbox`, and `BitStringLiteral.Signed` — the last has no `Is` prefix but is the same kind of boolean query, so it looked wrong left as *"Read-only property to access the signed"*.

**One decision I did not make for you:** `Name.HasPrefix` is the only `Has***` property. You wrote `Is***` specifically, so I left it on the `Read-only property to return ...` form. Both are boolean queries and it probably should follow the same rule — recorded in the findings file rather than changed unilaterally, since it is a wording call.

pyVHDLModel/__init__.py
Paebbels: Checked all of them, and written into the handoff.

There are 8 boolean-returning callables. Four already conformed; the four that did not are fixed:

| Callable | Was |
|---|---|
| `AllowBlackboxMixin.AllowBlackbox` | `:returns: If blackboxes are allowed.` |
| `Subprogram.IsPure` | one-liner, no `:returns:` |
| `Symbol.IsResolved` | one-liner, no `:returns:` |
| `BitStringLiteral.Signed` | one-liner, no `:returns:` |

All now start with ```` ``True``, if ... ````. Two of them are `Nullable[bool]` (`IsBlackbox`, `Signed`), so they document the `None` case as well — see the next comment.

pyVHDLModel/DesignUnit.py
Paebbels: You are right, and that was a genuine loss on my part — not a rewording. The property is `Nullable[bool]`, so `None` is a real outcome and I deleted the only sentence that said so.

Restored and extended to document all three:

```
Check if the component is a blackbox (:attr:`_isBlackBox`).

If components were not linked to matching entities, this property returns ``None``.

:returns: ``True``, if the component is a blackbox; ``False``, if it is not; ``None``, if components
          were not linked to entities yet.
```

I checked the rest of the diff for the same mistake. `BitStringLiteral.Signed` is the only other `Nullable[bool]`, and it had no `None` documentation to begin with — it now says ```` ``None``, if unspecified ````. The handoff rule now spells this out: for `Nullable[bool]`, document all three outcomes.

pyVHDLModel/__init__.py
Paebbels: "Indexed by" is better — "keyed by" is jargon that reads as a verb-noun mismatch. Replaced everywhere: 19 occurrences, 15 from this PR plus four in `Regions.py` that I had introduced in #148. No "keyed by" remains in the package.

pyVHDLModel/Name.py
Paebbels: Agreed, added to the findings file rather than done here.

The observation generalises: `NormalizedIdentifier`, `NormalizedIdentifiers` and `NormalizedLabel` all say "normalized" without ever explaining that VHDL identifiers are case-insensitive, so the normalized form is lower case and is what dictionary keys and name resolution use.

There is no glossary page in the documentation yet, so this needs one creating first — and the finding notes that is probably worth doing anyway for other terms that recur without definition: "declaration region", "declarative part", "design unit", "primary/secondary unit".

pyVHDLModel/__init__.py
Paebbels: Done — `Name.HasPrefix` now uses it too:

```
Check if the name has a prefix, i.e. :attr:`_prefix` is set.

This is true for all names except :class:`simple names <SimpleName>`.

:returns: ``True``, if the name has a prefix.
```

The handoff rule is widened accordingly — the row now reads *"boolean query — name starts with `Is***` or `Has***`"* rather than `Is***` alone — and the open question is removed from the findings file since it is settled.

Verified across the package: every boolean-query property uses the form, none left on the old lead. 428 passed, pyGHDL.dom cross-checked at 185 passed.

Paebbels:
Rechecked, and **227 of the 293 getter properties had no `:returns:`** — every one-liner from #149 plus the `Regions.py` ones from #148. All now have it (`c61464e`).

## Shaped by the return type, not boilerplate

| Return type | `:returns:` |
|---|---|
| `List[...]` | `List of ...` |
| `Dict[...]` | `Dictionary of ...` |
| `Tuple[...]` | `Tuple of ...` |
| `Nullable[...]` | ``The ..., or ``None`` if not set.`` |
| otherwise | `The ...` |

Where the dictionary key is knowable from the code that fills it, the line says so rather than leaving it vague — the `Regions.py` lookup tables and `DesignUnit`'s `Referenced*`/`Architectures` are all **indexed by normalized identifier**, and `Functions`/`Procedures` additionally note that each entry is a list of overloads.

Two dictionaries I did **not** claim a key for, because I did not verify it: `Namespace.SubNamespaces` and `RecordConstraint.Constraints` (the latter is keyed by `RecordElementSymbol`, not a string). They just say "Dictionary of ...".

Setters are excluded — they return nothing, and their doc-strings are invisible to Sphinx anyway.

## One gap this surfaced

**Ten properties have no return type annotation at all**, which is why they got the generic `The ...` form rather than a shaped one:

- `DesignUnit.py` — `Package.DeferredConstants`, `Package.Components`
- `Expression.py` — `UnaryExpression.Operand`, `BinaryExpression.LeftOperand`, `BinaryExpression.RightOperand`, `QualifiedExpression.Operand`, `QualifiedExpression.Subtype`, `AggregateElement.Expression`
- `Type.py` — `AccessType.DesignatedSubtype`, `FileType.DesignatedSubtype`

`CLAUDE.md` requires type hints on all functions and methods, so this is a real gap. Recorded in the findings file rather than fixed here — each needs the correct type determined from the assigning code rather than guessed, and it is a code change rather than documentation.

Verified from a fresh clone: **0** getter properties without `:returns:`. 428 passed, pyGHDL.dom cross-checked at 185 passed.

Paebbels:
Done (`c154bec`) — and it turned out to be 17, not 10.

## The ten properties

Each took the type **already declared for the field it exposes, in its own class**, so nothing was guessed:

| Property | Annotation |
|---|---|
| `Package.DeferredConstants` | `Dict[str, DeferredConstant]` |
| `Package.Components` | `Dict[str, 'Component']` |
| `UnaryExpression.Operand` | `ExpressionUnion` |
| `BinaryExpression.LeftOperand` / `RightOperand` | `ExpressionUnion` |
| `QualifiedExpression.Operand` | `ExpressionUnion` |
| `QualifiedExpression.Subtype` | `Symbol` |
| `AggregateElement.Expression` | `ExpressionUnion` |
| `AccessType.DesignatedSubtype` | `Symbol` |
| `FileType.DesignatedSubtype` | `Symbol` |

The two `Dict` ones now get the shaped doc-string the missing annotation had denied them — *"Dictionary of deferred constants, indexed by normalized identifier"* instead of the generic *"The deferred constants."* I verified the key from the code that fills it rather than assuming.

## Seven more found while checking

Scanning the whole package rather than only the ten showed `_IndexOtherDeclaredItem`, `Instantiate`, the `ParentNamespace` setter, `_AddLibraryClause`, `_AddPackageClause`, `_LinkSymbolsInExpression` and `_LinkItems` also had none. **None of them returns a value**, so all are `-> None` — unambiguous, no judgement needed.

## Verification

- **0** functions or methods in the package without a return annotation, confirmed from a fresh clone.
- Forced-annotation-evaluation sweep clean, so the new annotations also resolve on **Python 3.11–3.13**, where they are evaluated eagerly — that is the failure mode that bit #145/#146.
- 428 passed; pyGHDL.dom cross-checked at 185 passed.

> [!NOTE]
> Worth knowing for later: the annotation sweep only catches *unresolvable* annotations, not *absent* ones. Those need a separate AST scan, which is now noted in the findings entry.

PR #151, opened 2026-07-27
Document every class, with VHDL examples

Paebbels:
**All 429 classes now carry a doc-string with descriptive prose**, up from 245. **205 of them show the VHDL construct they abstract.** `interrogate` passes for the first time: **34.8% at the start of this work → 70.1%**, against a `fail-under` of 59.

## Conventions applied

- **Describes the class, not the initializer.** Constructor parameters stay on `__init__`.
- **`Represents a ...` summary**, then a paragraph describing the class through its properties, cross-referenced with `:data:`.
- **A VHDL example** in an `.. admonition:: Example`, underlining the part each property maps to.
- **Example variants** where a construct genuinely has different shapes — `Subtype` (no constraint / constraint / resolution function), `ArrayType` (one dimension / two unconstrained), `IfStatement` (bare `if` / with `elsif` and `else`).
- **Revisions**: pre-2019 no longer named — VHDL-2008 is the minimum supported and 87/93/2000/2002 are history. 17 such mentions removed. VHDL-2019 is still called out (mode views).

```
Represents a subtype declaration.

A subtype is a named entity (:data:`Identifier`, :data:`NormalizedIdentifier`) referencing a type
(:data:`Type`). Optionally, the subtype can be narrowed by a constraint (:data:`Range`) and/or
resolved by a resolution function (:data:`ResolutionFunction`).

.. admonition:: Example

   Without a constraint:

   .. code-block:: VHDL

      subtype byte is bit_vector;
      --      ^^^^                 <- Identifier
      --              ^^^^^^^^^^   <- Type
   ...
```

Classes that abstract no single construct — abstract bases, mixins, exceptions, the `BitStringBase` enum, the PSL classes — get prose describing their role instead of an example.

## Every snippet was verified with GHDL

Each example was analysed with `ghdl -a` before being written into a doc-string. That rejected **seven invalid drafts**: an access type as a signal, `o"240"` being nine bits not eight, a `wait` in a sensitized process, a shared variable of a non-protected type, `library work` inside a context declaration, a file type used before declaration, and a two-run label ambiguity.

## Underlines are computed, not typed

A generator derives caret columns and arrow alignment from the code line. Two bugs in it were caught by a checker rather than by reading — a target starting in the first columns collided with the `--` marker and produced negative padding, and targets on later lines of a multi-line snippet weren't found.

**The same checker found three misaligned underlines in *pre-existing* hand-written examples** (`ConditionalWaveform`, `ConditionalExpression`, `SelectedWaveform` in `Common.py`) — the ones this style was modelled on:

```
s <= '1' when cond else '0';
--   ^^^^^^^^^^^^         <- ...
```

The first run stopped one character short (`'1' when con`) and the second pointed at `' el'` — whitespace between tokens — rather than the final `'0'`. Regenerated, so every underline in the package now sits on real code.

## Verification

- 429/429 classes documented; 0 caret or arrow misalignments
- 0 structural RST problems (with Sphinx's `seealso`/`todo`/`pycode`/`|br|` registered — the ones bare docutils rejects are all pre-existing Sphinx constructs)
- 0 doc-string lines over 120 characters among those added
- 428 tests pass; pyGHDL.dom cross-checked at **185 passed, 0 failed**
- Verified from a fresh clone

### DIFF
--- a/pyVHDLModel/Base.py
+++ b/pyVHDLModel/Base.py
@@ -171,16 +171,16 @@ class NamedEntityMixin(metaclass=ExtendedType, mixin=True):
 
 	.. seealso::
 
-	   * :class:`Alias <pyVHDLModel.Declaration.Alias>`
 	   * :class:`Attribute <pyVHDLModel.Declaration.Attribute>`
-	   * :class:`Base type <pyVHDLModel.Type.BaseType>`
-	   * :class:`Component <pyVHDLModel.DesignUnit.Component>`
-	   * :class:`Default clock <pyVHDLModel.PSLModel.DefaultClock>`
+	   * :class:`Alias <pyVHDLModel.Declaration.Alias>`
 	   * :class:`Design unit <pyVHDLModel.DesignUnit.DesignUnit>`
-	   * :class:`Interface package <pyVHDLModel.Interface.InterfacePackage>`
-	   * :class:`Library <pyVHDLModel.Library>`
+	   * :class:`Component <pyVHDLModel.DesignUnit.Component>`
 	   * :class:`Mode view declaration <pyVHDLModel.Interface.ModeViewDeclaration>`
+	   * :class:`Interface package <pyVHDLModel.Interface.InterfacePackage>`
+	   * :class:`Default clock <pyVHDLModel.PSLModel.DefaultClock>`
 	   * :class:`Subprogram <pyVHDLModel.Subprogram.Subprogram>`
+	   * :class:`Base type <pyVHDLModel.Type.BaseType>`
+	   * :class:`Library <pyVHDLModel.Library>`
 	"""
 
 	_identifier:           str  #: The identifier of a model entity.
@@ -371,9 +371,9 @@ class LabeledEntityMixin(metaclass=ExtendedType, mixin=True):
 
 	.. seealso::
 
+	   * :class:`Statement <pyVHDLModel.Common.Statement>`
 	   * :class:`Concurrent block statement <pyVHDLModel.Concurrent.ConcurrentBlockStatement>`
 	   * :class:`Concurrent case <pyVHDLModel.Concurrent.ConcurrentCase>`
-	   * :class:`Statement <pyVHDLModel.Common.Statement>`
 	"""
 	_label:           Nullable[str]  #: The label of a model entity.
 	_normalizedLabel: Nullable[str]  #: The normalized (lower case) label of a model entity.
@@ -442,13 +442,13 @@ class ConditionalMixin(metaclass=ExtendedType, mixin=True):
 
 	.. seealso::
 
-	   * :class:`Assert statement mixin <pyVHDLModel.Base.AssertStatementMixin>`
 	   * :class:`Conditional branch mixin <pyVHDLModel.Base.ConditionalBranchMixin>`
-	   * :class:`Conditional expression <pyVHDLModel.Common.ConditionalExpression>`
+	   * :class:`Assert statement mixin <pyVHDLModel.Base.AssertStatementMixin>`
 	   * :class:`Conditional waveform <pyVHDLModel.Common.ConditionalWaveform>`
+	   * :class:`Conditional expression <pyVHDLModel.Common.ConditionalExpression>`
+	   * :class:`While loop statement <pyVHDLModel.Sequential.WhileLoopStatement>`
 	   * :class:`Loop control statement <pyVHDLModel.Sequential.LoopControlStatement>`
 	   * :class:`Wait statement <pyVHDLModel.Sequential.WaitStatement>`
-	   * :class:`While loop statement <pyVHDLModel.Sequential.WhileLoopStatement>`
 	"""
 
 	_condition: ExpressionUnion
@@ -497,8 +497,8 @@ class ConditionalBranchMixin(BranchMixin, ConditionalMixin, mixin=True):
 
 	.. seealso::
 
-	   * :class:`Elsif branch mixin <pyVHDLModel.Base.ElsifBranchMixin>`
 	   * :class:`If branch mixin <pyVHDLModel.Base.IfBranchMixin>`
+	   * :class:`Elsif branch mixin <pyVHDLModel.Base.ElsifBranchMixin>`
 	"""
 	def __init__(self, condition: ExpressionUnion) -> None:
 		super().__init__()
@@ -512,8 +512,8 @@ class IfBranchMixin(ConditionalBranchMixin, mixin=True):
 
 	.. seealso::
 
-	   * :class:`If branch <pyVHDLModel.Sequential.IfBranch>`
 	   * :class:`If generate branch <pyVHDLModel.Concurrent.IfGenerateBranch>`
+	   * :class:`If branch <pyVHDLModel.Sequential.IfBranch>`
 	"""
 
 
@@ -524,8 +524,8 @@ class ElsifBranchMixin(ConditionalBranchMixin, mixin=True):
 
 	.. seealso::
 
-	   * :class:`Elsif branch <pyVHDLModel.Sequential.ElsifBranch>`
 	   * :class:`Elsif generate branch <pyVHDLModel.Concurrent.ElsifGenerateBranch>`
+	   * :class:`Elsif branch <pyVHDLModel.Sequential.ElsifBranch>`
 	"""
 
 
@@ -536,8 +536,8 @@ class ElseBranchMixin(BranchMixin, mixin=True):
 
 	.. seealso::
 
-	   * :class:`Else branch <pyVHDLModel.Sequential.ElseBranch>`
 	   * :class:`Else generate branch <pyVHDLModel.Concurrent.ElseGenerateBranch>`
+	   * :class:`Else branch <pyVHDLModel.Sequential.ElseBranch>`
 	"""
 
 
@@ -631,11 +631,11 @@ class BaseCase(ModelEntity):
 
 	.. seealso::
 
-	   * :class:`Concurrent case <pyVHDLModel.Concurrent.ConcurrentCase>`
-	   * :class:`Others selected expression <pyVHDLModel.Common.OthersSelectedExpression>`
+	   * :class:`Selected waveform <pyVHDLModel.Common.SelectedWaveform>`
 	   * :class:`Others selected waveform <pyVHDLModel.Common.OthersSelectedWaveform>`
 	   * :class:`Selected expression <pyVHDLModel.Common.SelectedExpression>`
-	   * :class:`Selected waveform <pyVHDLModel.Common.SelectedWaveform>`
+	   * :class:`Others selected expression <pyVHDLModel.Common.OthersSelectedExpression>`
+	   * :class:`Concurrent case <pyVHDLModel.Concurrent.ConcurrentCase>`
 	   * :class:`Sequential case <pyVHDLModel.Sequential.SequentialCase>`
 	"""
 
@@ -647,9 +647,9 @@ class ChoicesMixin(metaclass=ExtendedType, mixin=True):
 
 	.. seealso::
 
-	   * :class:`Concurrent case <pyVHDLModel.Concurrent.ConcurrentCase>`
-	   * :class:`Selected expression <pyVHDLModel.Common.SelectedExpression>`
 	   * :class:`Selected waveform <pyVHDLModel.Common.SelectedWaveform>`
+	   * :class:`Selected expression <pyVHDLModel.Common.SelectedExpression>`
+	   * :class:`Concurrent case <pyVHDLModel.Concurrent.ConcurrentCase>`
 	   * :class:`Sequential case <pyVHDLModel.Sequential.SequentialCase>`
 	"""
 
@@ -682,8 +682,8 @@ class Range(ModelEntity):
 
 	.. seealso::
 
-	   * :class:`Range from name <pyVHDLModel.Base.RangeFromName>`
 	   * :class:`Simple range <pyVHDLModel.Base.SimpleRange>`
+	   * :class:`Range from name <pyVHDLModel.Base.RangeFromName>`
 	"""
 
 
--- a/pyVHDLModel/Base.py
+++ b/pyVHDLModel/Base.py
@@ -168,6 +168,19 @@ class NamedEntityMixin(metaclass=ExtendedType, mixin=True):
 
 	Protected variables :attr:`_identifier` and :attr:`_normalizedIdentifier` are available to derived classes as well as
 	two readonly properties :attr:`Identifier` and :attr:`NormalizedIdentifier` for public access.
+
+	.. seealso::
+
+	   * :class:`Alias <pyVHDLModel.Declaration.Alias>`
+	   * :class:`Attribute <pyVHDLModel.Declaration.Attribute>`
+	   * :class:`Base type <pyVHDLModel.Type.BaseType>`
+	   * :class:`Component <pyVHDLModel.DesignUnit.Component>`
+	   * :class:`Default clock <pyVHDLModel.PSLModel.DefaultClock>`
+	   * :class:`Design unit <pyVHDLModel.DesignUnit.DesignUnit>`
+	   * :class:`Interface package <pyVHDLModel.Interface.InterfacePackage>`
+	   * :class:`Library <pyVHDLModel.Library>`
+	   * :class:`Mode view declaration <pyVHDLModel.Interface.ModeViewDeclaration>`
+	   * :class:`Subprogram <pyVHDLModel.Subprogram.Subprogram>`
 	"""
 
 	_identifier:           str  #: The identifier of a model entity.
@@ -208,6 +221,10 @@ class OptionallyNamedEntityMixin(metaclass=ExtendedType, mixin=True):
 
 	Protected variables :attr:`_identifier` and :attr:`_normalizedIdentifier` are available to derived classes as well as
 	two readonly properties :attr:`Identifier` and :attr:`NormalizedIdentifier` for public access.
+
+	.. seealso::
+
+	   * :class:`Interface group <pyVHDLModel.Interface.InterfaceGroup>`
 	"""
 
 	_identifier:           Nullable[str]  #: The identifier of a model entity.
@@ -249,6 +266,12 @@ class MultipleNamedEntityMixin(metaclass=ExtendedType, mixin=True):
 
 	Protected variables :attr:`_identifiers` and :attr:`_normalizedIdentifiers` are available to derived classes as well
 	as two readonly properties :attr:`Identifiers` and :attr:`NormalizedIdentifiers` for public access.
+
+	.. seealso::
+
+	   * :class:`Mode view element <pyVHDLModel.Interface.ModeViewElement>`
+	   * :class:`Obj <pyVHDLModel.Object.Obj>`
+	   * :class:`Record type element <pyVHDLModel.Type.RecordTypeElement>`
 	"""
 
 	_identifiers:           Tuple[str]  #: A list of identifiers.
@@ -345,6 +368,12 @@ class LabeledEntityMixin(metaclass=ExtendedType, mixin=True):
 
 	protected variables :attr:`_label` and :attr:`_normalizedLabel` are available to derived classes as well as two
 	readonly properties :attr:`Label` and :attr:`NormalizedLabel` for public access.
+
+	.. seealso::
+
+	   * :class:`Concurrent block statement <pyVHDLModel.Concurrent.ConcurrentBlockStatement>`
+	   * :class:`Concurrent case <pyVHDLModel.Concurrent.ConcurrentCase>`
+	   * :class:`Statement <pyVHDLModel.Common.Statement>`
 	"""
 	_label:           Nullable[str]  #: The label of a model entity.
 	_normalizedLabel: Nullable[str]  #: The normalized (lower case) label of a model entity.
@@ -408,7 +437,19 @@ class DocumentedEntityMixin(metaclass=ExtendedType, mixin=True):
 
 @export
 class ConditionalMixin(metaclass=ExtendedType, mixin=True):
-	"""A ``ConditionalMixin`` is a mixin-class for all statements with a condition."""
+	"""
+	A ``ConditionalMixin`` is a mixin-class for all statements with a condition.
+
+	.. seealso::
+
+	   * :class:`Assert statement mixin <pyVHDLModel.Base.AssertStatementMixin>`
+	   * :class:`Conditional branch mixin <pyVHDLModel.Base.ConditionalBranchMixin>`
+	   * :class:`Conditional expression <pyVHDLModel.Common.ConditionalExpression>`
+	   * :class:`Conditional waveform <pyVHDLModel.Common.ConditionalWaveform>`
+	   * :class:`Loop control statement <pyVHDLModel.Sequential.LoopControlStatement>`
+	   * :class:`Wait statement <pyVHDLModel.Sequential.WaitStatement>`
+	   * :class:`While loop statement <pyVHDLModel.Sequential.WhileLoopStatement>`
+	"""
 
 	_condition: ExpressionUnion
 
@@ -436,7 +477,14 @@ class ConditionalMixin(metaclass=ExtendedType, mixin=True):
 
 @export
 class BranchMixin(metaclass=ExtendedType, mixin=True):
-	"""A ``BranchMixin`` is a mixin-class for all statements with branches."""
+	"""
+	A ``BranchMixin`` is a mixin-class for all statements with branches.
+
+	.. seealso::
+
+	   * :class:`Conditional branch mixin <pyVHDLModel.Base.ConditionalBranchMixin>`
+	   * :class:`Else branch mixin <pyVHDLModel.Base.ElseBranchMixin>`
+	"""
 
 	def __init__(self) -> None:
 		pass
@@ -444,7 +492,14 @@ class BranchMixin(metaclass=ExtendedType, mixin=True):
 
 @export
 class ConditionalBranchMixin(BranchMixin, ConditionalMixin, mixin=True):
-	"""A ``BaseBranch`` is a mixin-class for all branch statements with a condition."""
+	"""
+	A ``BaseBranch`` is a mixin-class for all branch statements with a condition.
+
+	.. seealso::
+
+	   * :class:`Elsif branch mixin <pyVHDLModel.Base.ElsifBranchMixin>`
+	   * :class:`If branch mixin <pyVHDLModel.Base.IfBranchMixin>`
+	"""
 	def __init__(self, condition: ExpressionUnion) -> None:
 		super().__init__()
 		ConditionalMixin.__init__(self, condition)
@@ -452,22 +507,50 @@ class ConditionalBranchMixin(BranchMixin, ConditionalMixin, mixin=True):
 
 @export
 class IfBranchMixin(ConditionalBranchMixin, mixin=True):
-	"""A ``BaseIfBranch`` is a mixin-class for all if-branches."""
+	"""
+	A ``BaseIfBranch`` is a mixin-class for all if-branches.
+
+	.. seealso::
+
+	   * :class:`If branch <pyVHDLModel.Sequential.IfBranch>`
+	   * :class:`If generate branch <pyVHDLModel.Concurrent.IfGenerateBranch>`
+	"""
 
 
 @export
 class ElsifBranchMixin(ConditionalBranchMixin, mixin=True):
-	"""A ``BaseElsifBranch`` is a mixin-class for all elsif-branches."""
+	"""
+	A ``BaseElsifBranch`` is a mixin-class for all elsif-branches.
+
+	.. seealso::
+
+	   * :class:`Elsif branch <pyVHDLModel.Sequential.ElsifBranch>`
+	   * :class:`Elsif generate branch <pyVHDLModel.Concurrent.ElsifGenerateBranch>`
+	"""
 
 
 @export
 class ElseBranchMixin(BranchMixin, mixin=True):
-	"""A ``BaseElseBranch`` is a mixin-class for all else-branches."""
+	"""
+	A ``BaseElseBranch`` is a mixin-class for all else-branches.
+
+	.. seealso::
+
+	   * :class:`Else branch <pyVHDLModel.Sequential.ElseBranch>`
+	   * :class:`Else generate branch <pyVHDLModel.Concurrent.ElseGenerateBranch>`
+	"""
 
 
 @export
 class ReportStatementMixin(metaclass=ExtendedType, mixin=True):
-	"""A ``MixinReportStatement`` is a mixin-class for all report and assert statements."""
+	"""
+	A ``MixinReportStatement`` is a mixin-class for all report and assert statements.
+
+	.. seealso::
+
+	   * :class:`Assert statement mixin <pyVHDLModel.Base.AssertStatementMixin>`
+	   * :class:`Sequential report statement <pyVHDLModel.Sequential.SequentialReportStatement>`
+	"""
 
 	_message:  Nullable[ExpressionUnion]
 	_severity: Nullable[ExpressionUnion]
@@ -502,7 +585,14 @@ class ReportStatementMixin(metaclass=ExtendedType, mixin=True):
 
 @export
 class AssertStatementMixin(ReportStatementMixin, ConditionalMixin, mixin=True):
-	"""A ``MixinAssertStatement`` is a mixin-class for all assert statements."""
+	"""
+	A ``MixinAssertStatement`` is a mixin-class for all assert statements.
+
+	.. seealso::
+
+	   * :class:`Concurrent assert statement <pyVHDLModel.Concurrent.ConcurrentAssertStatement>`
+	   * :class:`Sequential assert statement <pyVHDLModel.Sequential.SequentialAssertStatement>`
+	"""
 
 	def __init__(self, condition: ExpressionUnion, message: Nullable[ExpressionUnion] = None, severity: Nullable[ExpressionUnion] = None) -> None:
 		super().__init__(message, severity)
@@ -510,7 +600,13 @@ class AssertStatementMixin(ReportStatementMixin, ConditionalMixin, mixin=True):
 
 
 class BlockStatementMixin(metaclass=ExtendedType, mixin=True):
-	"""A ``BlockStatement`` is a mixin-class for all block statements."""
+	"""
+	A ``BlockStatement`` is a mixin-class for all block statements.
+
+	.. seealso::
+
+	   * :class:`Concurrent block statement <pyVHDLModel.Concurrent.ConcurrentBlockStatement>`
+	"""
 
 	def __init__(self) -> None:
 		pass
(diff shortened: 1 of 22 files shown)


Paebbels:
Thanks — that corrects something I got wrong in the PR description.

I reported one remaining ReST problem (`Undefined substitution referenced: "rarr"` in `DesignUnit.Context`) and said the rejected constructs were "pre-existing Sphinx constructs". That was right about the cause but wrong about the diagnosis: **`|rarr|` is not a Sphinx built-in, it comes from docutils' `<isonum.txt>`**, which `doc/prolog.inc` includes. My validator simply wasn't using the project's prolog.

Reading `doc/prolog.inc` properly, it provides:

- `<isonum.txt>` and `<mmlalias.txt>` — where `|rarr|` and friends come from
- `|br|` and `|hr|` as raw HTML
- the roles `pycode`, `bolditalic`, `underline`, `strike`, `xlarge`, `red`, `green`, `blue`, `purple`, `deletion`, `addition`

So `:pycode:` was legitimate too — I had stubbed it as an unknown role, which masked rather than explained it.

Re-validated with the real prolog prepended, stubbing only the genuinely Sphinx-only pieces (`:class:`/`:attr:`/`:data:`/... domain roles, and `seealso`/`todo`/`versionadded`/...):

```
ReST problems: 0
```

Across **every** doc-string in the package, not just the ones this PR touches. I also self-tested the checker against a deliberately broken doc-string to confirm a zero means something — it catches a malformed `code-block` and a genuinely undefined substitution.

One practical note recorded in the handoff: the prolog pulls in include files, so it must be parsed **once per file** with that file's doc-strings concatenated and error line numbers mapped back to the owning class. Parsing it once per doc-string takes minutes and times out.

No code changes — the package was already correct here; only my verification was wrong.

pyVHDLModel/Concurrent.py
```diff
@@ -363,6 +363,17 @@ def SensitivityList(self) -> List[Name]:
...
+	"""
+	Represents a concurrent procedure call.
+
+	.. admonition:: Example
+
+	   .. code-block:: VHDL
+
+	      proc_lbl : proc(clk, open);
+	    --^^^^^^^^                      <- Label
+	    --           ^^^^^^^^^^^^^^^    <- the call
```
Paebbels:
```suggestion
	      proc_lbl : proc(clk, open);
	      --^^^^^^                      <- Label
	      --         ^^^^^^^^^^^^^^^    <- the call
```
The VHDL comment markers where outside of the code-block. This would cause a syntax error. Maybe you generator checker uses a different setting for tab-width causing this problem. According to `.editorconfig`, a tab is 2 spaces wide for this repository.

pyVHDLModel/Concurrent.py
```diff
@@ -376,6 +387,25 @@ def __init__(
...
+
+	   .. code-block:: VHDL
+
+	      blk : block
+	    --^^^                            <- Label
+	        port (bp : in bit);
+	      --      ^^^^^^^^^^^            <- PortItems
+	        signal inner : bit := '0';
+	      --^^^^^^^^^^^^^^^^^^^^^^^^^^   <- DeclaredItems
+	      begin
```
Paebbels: Where are the statements ?

Paebbels:
Added `.. seealso::` cross-links (`f64d026`). **180 classes now carry one, holding 488 links — all verified to resolve, none duplicated.**

## Two kinds of link

**Derived classes**, generated from the hierarchy for the **121** classes with 1–10 direct subclasses. As you said, Sphinx lists base-classes automatically but not subclasses, so this is the direction worth writing down. Classes with more than ten (`ModelEntity` 37, `Symbol` 21, `SequentialStatement` 16, ...) are skipped — a list that long is noise, not navigation.

This alone gives your first example for free: `DesignUnit` → `PrimaryUnit`/`SecondaryUnit` → `Entity`/`Package`/`Context`/`Configuration`, since each step is a subclass relation.

**Curated cross-links** for relationships the hierarchy cannot express (**54** classes):

| | |
|---|---|
| generate statements | if ↔ case ↔ for |
| loop statements | endless ↔ for ↔ while |
| concurrent ↔ sequential | each assignment, assert and procedure call to its counterpart, both ways |
| `WaveformElement` | → the three places a waveform is held (`WaveformMixin`, `ConditionalWaveform`, `SelectedWaveform`), and back |
| design units | entity ↔ architecture, package ↔ package body, entity ↔ component ↔ configuration |
| types | protected type ↔ its body, record type ↔ record element ↔ `RecordElementSymbol` |
| mode views | declaration ↔ the ports and parameters that use it |
| regions | each declaration region ↔ the other ↔ `Namespace` |

Also cross-linked `ForLoopStatement` → `ForGenerateStatement`, `IfStatement` → `IfGenerateStatement` and `CaseStatement` → `CaseGenerateStatement`, since the sequential and concurrent forms of the same construct are easy to confuse.

Base-classes and mixins are deliberately **not** listed, per your point that Sphinx shows them already.

## Two defects caught before pushing

- **Duplicates.** Four classes (`PrimaryUnit` among them) already linked their subclasses using the ``:class:`~pkg.Cls``` short form. My first pass de-duplicated by bullet *text*, so it appended the same classes again in long form. It now de-duplicates by **target**, and I verified 0 duplicated targets.
- **`Psl primary unit`.** My label generator ran `.capitalize()` over the whole string, flattening the acronym. Fixed to preserve upper-case words: `PSL primary unit`, `VHDL version`.

## Verification

- 180 seealso blocks, 488 links, **0 dangling targets**, **0 duplicated targets** — checked from a fresh clone
- 0 ReST problems (validated with `doc/prolog.inc`), 0 caret misalignments, 0 lines over 120 characters
- 428 tests pass; pyGHDL.dom cross-checked at **185 passed, 0 failed**

pyVHDLModel/Concurrent.py
```diff
@@ -376,6 +387,25 @@ def __init__(
 
 @export
 class ConcurrentBlockStatement(ConcurrentStatement, BlockStatementMixin, LabeledEntityMixin, WithPortsMixin, ConcurrentDeclarationRegionMixin, ConcurrentStatementsMixin, DocumentedEntityMixin, AllowBlackboxMixin):
+	"""
+	Represents a block statement.
+
+	A block groups concurrent statements and may declare its own items. It can also have a port
+	clause (:data:`PortItems`), which makes it a hierarchy level of its own.
```
Paebbels: the sub-sentence indicates the hierarchy is linked to the ports, but hierarchy is always created independent of port items.

pyVHDLModel/Base.py
```diff
@@ -436,38 +477,80 @@ def Condition(self) -> ExpressionUnion:
...
 @export
 class ConditionalBranchMixin(BranchMixin, ConditionalMixin, mixin=True):
-	"""A ``BaseBranch`` is a mixin-class for all branch statements with a condition."""
+	"""
+	A ``BaseBranch`` is a mixin-class for all branch statements with a condition.
+
+	.. seealso::
+
+	   * :class:`Elsif branch mixin <pyVHDLModel.Base.ElsifBranchMixin>`
+	   * :class:`If branch mixin <pyVHDLModel.Base.IfBranchMixin>`
```
Paebbels:
```suggestion
	   * :class:`If branch mixin <pyVHDLModel.Base.IfBranchMixin>`
	   * :class:`Elsif branch mixin <pyVHDLModel.Base.ElsifBranchMixin>`
```

pyVHDLModel/Common.py
```diff
@@ -138,7 +172,16 @@ def ParameterAssociationItems(self) -> List[ParameterAssociationItem]:
 
 @export
 class AssignmentMixin(metaclass=ExtendedType, mixin=True):
-	"""A mixin-class for all assignment statements."""
+	"""
+	A mixin-class for all assignment statements.
+
+	.. seealso::
+
+	   * :class:`~pyVHDLModel.Sequential.SequentialConditionalVariableAssignment`
```
Paebbels: why does this line have a different format not using <...>?

pyVHDLModel/Common.py
```diff
@@ -345,7 +456,18 @@ def __init__(
...
+	"""
+	Represents the ``others`` alternative of a selected signal assignment.
+
+	It supplies the waveform (:data:`Waveform`) for every choice not named explicitly.
+
+	.. admonition:: Example
+
+	   .. code-block:: VHDL
+
+	      with sel select s <= '1' when 0, '0' when others;
```
Paebbels: The first 0 (without `'`) is a condition, thus it must evaluate to a boolean. 0 (integer) isn't a good example. Use e.g. `sel = '0'`.

pyVHDLModel/Concurrent.py
```diff
@@ -574,13 +637,15 @@ def __init__(
...
+	A generate statement replicates or conditionally elaborates concurrent statements.
 
 	.. seealso::
 
-	   * :class:`If...generate statement <pyVHDLModel.Concurrent.IfGenerateStatement>`
-	   * :class:`Case...generate statement <pyVHDLModel.Concurrent.CaseGenerateStatement>`
-	   * :class:`For...generate statement <pyVHDLModel.Concurrent.ForGenerateStatement>`
+	   * :class:`Case generate statement <pyVHDLModel.Concurrent.CaseGenerateStatement>`
+	   * :class:`For generate statement <pyVHDLModel.Concurrent.ForGenerateStatement>`
+	   * :class:`If generate statement <pyVHDLModel.Concurrent.IfGenerateStatement>`
```
Paebbels:
```suggestion
	   * :class:`If generate statement <pyVHDLModel.Concurrent.IfGenerateStatement>`
	   * :class:`Case generate statement <pyVHDLModel.Concurrent.CaseGenerateStatement>`
	   * :class:`For generate statement <pyVHDLModel.Concurrent.ForGenerateStatement>`
```

pyVHDLModel/DesignUnit.py
```diff
@@ -351,6 +359,7 @@ class PrimaryUnit(DesignUnit):
 	   * :class:`~pyVHDLModel.DesignUnit.Entity`
 	   * :class:`~pyVHDLModel.DesignUnit.Package`
 	   * :class:`~pyVHDLModel.DesignUnit.Configuration`
```
Paebbels: adjust the format.

pyVHDLModel/DesignUnit.py
```diff
@@ -351,6 +359,7 @@ class PrimaryUnit(DesignUnit):
...
+	   * :class:`PSL primary unit <pyVHDLModel.PSLModel.PSLPrimaryUnit>`
```
Paebbels: should we mark this as unsupported?

pyVHDLModel/Exception.py
```diff
@@ -43,27 +43,62 @@
...
+	"""
+	Raised when a language construct is recognised but not yet modelled.
+	"""
 	pass
 
 
 @export
 class VHDLModelCriticalWarning(Warning):
+	"""
+	Base-class for warnings that likely indicate a defect in the model or its input.
```
Paebbels:
```suggestion
	Base-class for critical warnings that likely indicate a defect in the model or its input.
```

pyVHDLModel/Subprogram.py
```diff
@@ -147,6 +159,28 @@ def IndexDeclaredItems(self) -> None:
...
+	Represents a procedure.
+
+	Unlike a function, a procedure returns no value.
+
+	.. admonition:: Example
+
+	   .. code-block:: VHDL
+
+	      procedure proc(signal s : in bit) is
+	    --^^^^                                   <- Identifier
```
Paebbels: proc triggered at the wrong position because of keyword and name overlap.

pyVHDLModel/Subprogram.py
```diff
@@ -147,6 +159,28 @@ def IndexDeclaredItems(self) -> None:
...
+	    --                      ^^^^^^^^^^       <- ParameterItems
+	      begin
+	      end procedure;
```
Paebbels, line 181: where are the declared items and statements?

pyVHDLModel/Subprogram.py
```diff
@@ -162,6 +196,30 @@ def __init__(
...
+	.. admonition:: Example
+
+	   .. code-block:: VHDL
+
+	      function fun(constant c : in integer) return integer is
+	    --^^^                                                       <- Identifier
+	    --                      ^^^^^^^^^^^^^^                      <- ParameterItems
+	    --                             ^^^^^^^                      <- ReturnType
+	      begin
+	      end function;
```
Paebbels: same remarks as for Procedure. in addition, because parameter type and return overlap, the `^^^^` is positions wrongly. maybe the input type is a positive and output an integer.

pyVHDLModel/Concurrent.py
Paebbels: Corrected. The prose now reads that a block statement always forms a hierarchy level, and that — independently of that — it may also carry a generic and/or port clause. The example gained a `Statements` marker too.

pyVHDLModel/Base.py
Paebbels: Applied verbatim.

pyVHDLModel/Common.py
Paebbels: Oversight on my side — that bullet was left in the short `~` form. All `seealso` bullets across the PR are now in the labelled `:class:`Label <target>`` format (494 links in 180 blocks, verified by a script that also checks for dangling and duplicated targets). The only blocks still using the short form are the pre-existing ones in `IEEE.py`/`STD.py`, where the class name already is the best label.

pyVHDLModel/Concurrent.py
Paebbels: Applied verbatim.

pyVHDLModel/DesignUnit.py
Paebbels: Adjusted — the whole block is labelled now. I also made the nested list in the `DesignUnit` base-class consistent with it.

pyVHDLModel/Exception.py
Paebbels: Applied verbatim.

pyVHDLModel/DesignUnit.py
Paebbels: Yes — marked as `(PSL is not supported)`.

pyVHDLModel/Subprogram.py
Paebbels: Fixed. The generator now matches a specific occurrence instead of the first substring match, so `proc` is no longer hit inside the `procedure` keyword. I audited the whole PR for this class of error and found 31 caret runs sitting mid-identifier (a few of them pre-existing); all 31 are corrected and a check script now reports 0.

pyVHDLModel/Subprogram.py
Paebbels: Added — both `Procedure` and `Function` examples now show the declarative part and the statement part with their own markers.

pyVHDLModel/Subprogram.py
Paebbels: Fixed the same way. To remove the overlap I made the parameter a `positive` and the return type an `integer`, so the two caret runs are unambiguous.

pyVHDLModel/Common.py
Paebbels: Agreed that the bare `0` was wrong, but I deviated slightly from your suggestion — please correct me if I read the construct wrong.

That example is a *selected* assignment (`with sel select ...`), so the thing after `when` is a **choice**, not a condition; it has to be of the type of the selector expression. `sel = 0` would be a boolean expression and is not a legal choice there. I used a bit literal instead:

```vhdl
with sel select s <= '1' when '0', '0' when others;
```

The *conditional* assignments (`ConditionalWaveform`, `ConditionalExpression`) are the ones taking a real condition, and those use `cond` in the examples.

I verified all four assignment examples against the GHDL analyzer (`ghdl -a --std=08`, `signal sel, s : bit`) and they analyze clean.

Paebbels:
All 13 review comments are addressed in 811ac5f.

Validation run on the branch:

| Check | Result |
|---|---|
| Caret/arrow alignment (all example blocks) | 0 problems |
| `seealso` links | 494 links / 180 blocks, 0 dangling, 0 duplicated, 0 mixed-format |
| `interrogate` | 70.1% (PASSED, minimum 59.0%) |
| GHDL analyzer on the changed examples | clean |

Two notes:

* On the selected-assignment example I did not take the suggestion literally — `sel = '0'` is a boolean expression and cannot be a choice in `with ... select`; I used the bit literal `when '0'` instead. Details in the thread.
* While normalizing the `seealso` blocks I had flattened the pre-existing nested list in `DesignUnit`. That is restored to its original two-level structure, now with labelled bullets throughout.

pyVHDLModel/Concurrent.py
```diff
@@ -376,6 +416,33 @@ def __init__(
...
+
+	.. admonition:: Example
+
+	   .. code-block:: VHDL
+
+	        blk : block
+	      --^^^                            <- Label
+	          port (bp : in bit);
+	      --        ^^^^^^^^^^^            <- PortItems
+	          port map (bp => clk);
```
Paebbels: I think we have no field for the port map items, right? also a marking is missing. Put into findings if this feature is missing.

pyVHDLModel/Type.py
```diff
@@ -398,6 +673,29 @@ def Methods(self) -> List[Union['Procedure', 'Function']]:
...
+	A protected type body implements the methods (:data:`Methods`) declared by the
+	:class:`ProtectedType` of the same identifier (:data:`Identifier`).
+
+	.. admonition:: Example
+
+	   .. code-block:: VHDL
+
+	      type counter is protected body
+	      --   ^^^^^^^                        <- Identifier
+	        variable count : natural := 0;
```
Paebbels, line 704: Doesn't it need to point out declared elements?

pyVHDLModel/Type.py
```diff
@@ -375,6 +629,27 @@ def __str__(self) -> str:
...
+	A protected type is a named entity (:data:`Identifier`) exposing only its methods
+	(:data:`Methods`). The implementation lives in a separate :class:`ProtectedTypeBody`.
+
+	.. admonition:: Example
+
+	   .. code-block:: VHDL
+
+	      type counter is protected
+	      --   ^^^^^^^                             <- Identifier
+	        procedure increment;                   -- Methods
```
Paebbels: it should also point out declared elements.

pyVHDLModel/Sequential.py
```diff
@@ -81,6 +97,20 @@ def Statements(self) -> List[SequentialStatement]:
...
 @export
 class SequentialProcedureCall(SequentialStatement, ProcedureCallMixin):
+	"""
+	Represents a procedure call as a sequential statement.
+
+	.. admonition:: Example
+
+	   .. code-block:: VHDL
+
+	        log("hello");
```
Paebbels: isn't there also an optional label in general for all sequential statements?

pyVHDLModel/Sequential.py
```diff
@@ -94,20 +124,53 @@ def __init__(
...
 class SequentialSimpleSignalAssignment(SequentialSignalAssignment, WaveformMixin):
+	"""
+	Represents a simple sequential signal assignment.
+
+	.. admonition:: Example
+
+	   .. code-block:: VHDL
+
+	      s <= '1';
+	      --^         <- Target
```
Paebbels: `^` doesn't point to `s`

pyVHDLModel/Sequential.py
```diff
@@ -94,20 +124,53 @@ def __init__(
...
 class SequentialVariableAssignment(SequentialStatement, VariableAssignmentMixin):
+	"""
+	Represents a simple sequential variable assignment.
+
+	.. admonition:: Example
+
+	   .. code-block:: VHDL
+
+	      v := '1';
+	      --^         <- Target
```
Paebbels: `^` doesn't point to `v`.

pyVHDLModel/Sequential.py
```diff
@@ -116,11 +179,22 @@ def __init__(self, target: VariableSymbol, expression: ExpressionUnion, label: N
...
+
 	.. admonition:: Example
 
 	   .. code-block:: VHDL
 
-	      v := '1' when cond1 else '0' when cond2 else 'Z';
+	      v := '1' when sel = 0 else '0';
+	      --^                               <- Target
+	      --   ^^^^^^^^^^^^^^^^             <- first branch
+	      --                         ^^^    <- final branch
```
Paebbels: why does it use the terms first and final, the model has a list, but no individual class fields.

pyVHDLModel/Sequential.py
```diff
@@ -175,11 +259,19 @@ def __init__(
...
+	Represents a selected sequential variable assignment.
+
 	.. admonition:: Example
 
 	   .. code-block:: VHDL
 
-	      with sel select v := '1' when 0, '0' when others;
+	      with sel select v := '1' when '0', '0' when others;
+	      --   ^^^                                              <- the selector
+	      --                   ^^^^^^^^^^^^                     <- first alternative
```
Paebbels: why does it use the terms first, the model has a list, but no individual class fields.

pyVHDLModel/Sequential.py
```diff
@@ -294,27 +451,84 @@ def __init__(self, statements: Nullable[Iterable[SequentialStatement]] = None, p
...
 class IfBranch(Branch, IfBranchMixin):
+	"""
+	Represents the ``if`` branch of an if statement.
+
+	.. admonition:: Example
+
+	   .. code-block:: VHDL
+
+	      if sel = 0 then
+	      -- ^^^^^^^        <- Condition
```
Paebbels: where are the statements?

pyVHDLModel/Sequential.py
```diff
@@ -294,27 +451,84 @@ def __init__(self, statements: Nullable[Iterable[SequentialStatement]] = None, p
...
 class ElsifBranch(Branch, ElsifBranchMixin):
+	"""
+	Represents an ``elsif`` branch of an if statement.
+
+	.. admonition:: Example
+
+	   .. code-block:: VHDL
+
+	      elsif sel = 1 then
+	      --    ^^^^^^^        <- Condition
```
Paebbels: where are the statements?

pyVHDLModel/Sequential.py
```diff
@@ -294,27 +451,84 @@ def __init__(self, statements: Nullable[Iterable[SequentialStatement]] = None, p
...
 		super().__init__(statements, parent)
 		ElsifBranchMixin.__init__(self, condition)
 
 
 @export
 class ElseBranch(Branch, ElseBranchMixin):
+	"""
+	Represents the ``else`` branch of an if statement.
+
+	Unlike the other branches, an else branch has no condition.
```
Paebbels: where is the example to show the statements.

maybe, in all 3 cases (if, elsif, else) the full statement should be shown and then the marking should be done in 2D:
1) show what part of the if-statement we're focusing on (if, elsif, else) (like a big bracket)   
2) the individual parts like condition, statements, etc.

pyVHDLModel/Sequential.py
```diff
@@ -573,21 +941,67 @@ def LoopReference(self) -> LoopStatement:
...
+	Represents a ``next`` statement.
+
+	It skips to the next iteration of the enclosing loop.
+
+	.. admonition:: Example
+
+	   .. code-block:: VHDL
+
+	      next when k = 1;
+	      --        ^^^^^    <- Condition
```
Paebbels: general question should we mark optional parts like `Condition (optional)` or `optional Condition`?

pyVHDLModel/Sequential.py
```diff
@@ -614,6 +1028,20 @@ def ReturnValue(self) -> Nullable[ExpressionUnion]:
...
+	Represents a wait statement.
+
+	A wait statement may name a sensitivity list (:data:`SensitivityList`), a condition
+	(:data:`Condition`) and a timeout (:data:`Timeout`); all three are optional.
+
+	.. admonition:: Example
+
+	   .. code-block:: VHDL
+
+	      wait until clk = '1' for 10 ns;
```
Paebbels: Use `clock`, not `clk`. I think you have 2 other usage places like a `proc(clk, ..)` or so.

pyVHDLModel/Expression.py
```diff
@@ -389,61 +623,207 @@ def __str__(self) -> str:
...
+	"""
+	Represents a logical inversion expression (``not``).
+
+	The operand is available as :data:`Operand`.
+
+	.. admonition:: Example
+
+	   .. code-block:: VHDL
+
+	      bres := not boperand;
```
Paebbels: just go with `operand`, no `b` or `v`, because operations can be overloaded.

pyVHDLModel/Expression.py
```diff
@@ -389,61 +623,207 @@ def __str__(self) -> str:
...
+
+	A type conversion converts its operand (:data:`Operand`) to the target subtype
+	(:data:`TargetSubtype`). Unlike every other :class:`UnaryExpression`, its "operator" is the target
+	type name itself rather than a fixed string, so it carries its own subtype and renders itself.
+
+	.. admonition:: Example
+
+	   .. code-block:: VHDL
+
+	      res := integer(realval);
```
Paebbels: replace `realval` by `val` to stay generic.

pyVHDLModel/Expression.py
```diff
@@ -532,223 +949,823 @@ def Direction(self) -> Direction:
...
+	"""
+	Represents an exponentiation expression (``**``).
+
+	Both operands are available as :data:`LeftOperand` and :data:`RightOperand`.
+
+	.. admonition:: Example
+
+	   .. code-block:: VHDL
+
+	      res := lhs ** 2;
```
Paebbels: `2` should be `rhs`.

pyVHDLModel/Expression.py
```diff
@@ -532,223 +949,823 @@ def Direction(self) -> Direction:
...
+	"""
+	Represents a logical ``and`` expression.
+
+	Both operands are available as :data:`LeftOperand` and :data:`RightOperand`.
+
+	.. admonition:: Example
+
+	   .. code-block:: VHDL
+
+	      bres := blhs and brhs;
```
Paebbels: again no `b` prefix because operators can be overloaded.

pyVHDLModel/Expression.py
```diff
@@ -532,223 +949,823 @@ def Direction(self) -> Direction:
...
+	Represents a matching inequality expression (``?/=``).
+
+	Unlike ``/=``, a matching operator returns a ``bit``/``std_ulogic``.
+	Both operands are available as :data:`LeftOperand` and :data:`RightOperand`.
+
+	.. admonition:: Example
+
+	   .. code-block:: VHDL
+
+	      sres := vlhs ?/= vrhs;
```
Paebbels: no `v` prefix.

pyVHDLModel/Concurrent.py
Paebbels: Correct - there is no field. `ConcurrentBlockStatement` has `WithPortsMixin` (the block's *port clause*), but no `_genericAssociationItems`/`_portAssociationItems`, so the *port map aspect* is dropped entirely. Verified that the construct is legal (`ghdl -a --std=08` accepts a block with `generic`/`generic map`/`port`/`port map`).

Written up in the findings file. Note that `Instantiation` (`Concurrent.py:149`) already carries exactly these two fields and marks both `# TODO: extract to mixin` - so that extraction and this gap are one piece of work. I also added a `.. note::` to the class doc-string saying the port map aspect is not represented yet, so the missing marker is explained rather than just absent.

pyVHDLModel/Type.py
Paebbels: Yes. Both protected-type examples now mark the declared items with real markers instead of a trailing `-- Methods` comment.

For the *body* there is a wrinkle worth flagging: `ProtectedTypeBody.__init__` takes `declaredItems` but stores all of them into `_methods`, so `variable count` in the example would end up in `Methods` today. Rather than draw a marker that documents that as intended, I marked only the procedure and added a note that non-subprogram declarations are currently conflated with the methods. This is the `_methods` untangling already tracked in the findings file.

pyVHDLModel/Type.py
Paebbels: Done - `Methods[0]` and `Methods[1]` are marked now.

pyVHDLModel/Sequential.py
Paebbels: Yes - `Statement` derives from `LabeledEntityMixin`, so every sequential (and concurrent) statement can carry an optional label. Verified with the analyzer that all of them accept one.

I added the label to the `SequentialProcedureCall` and `ConcurrentProcedureCall` examples and said so in the prose. I deliberately did *not* add it to all ~25 other statement examples - it would double their size and push the distinguishing field out of view. Tell me if you would rather have it everywhere, or documented once on the `SequentialStatement` base-class instead.

pyVHDLModel/Sequential.py
Paebbels: Fixed. The cause was general, not local: when a target sits in column 0 the code has to be shifted right so the `--` marker does not sit on top of it. That shift was not applied for these, so the caret landed two columns right - on the `<` of `<=`.

I searched for the pattern instead of just fixing the two you found: **six** examples were affected (`SequentialSimpleSignalAssignment`, `SequentialVariableAssignment`, `SequentialConditionalVariableAssignment`, `SequentialConditionalSignalAssignment`, `SignalForceAssignment`, `SignalReleaseAssignment`). All six are regenerated.

My caret checker had missed them because it only verified that a caret sits on a non-space character - and `<` qualifies. It now also rejects any caret run made up purely of operator/punctuation characters, which is what catches this class of error.

pyVHDLModel/Sequential.py
Paebbels: Same root cause, fixed in the same sweep - see the reply on the signal assignment above.

pyVHDLModel/Sequential.py
Paebbels: Agreed - the wording implied fields that do not exist. The markers now name the list and the element index, e.g. `<- ConditionalExpressions[0]` / `<- ConditionalExpressions[1]`, and the prose says explicitly that the model holds them in a list with no distinct field per alternative.

I applied the same convention everywhere it was wrong: `ConditionalWaveforms`, `SelectedExpressions`, `SelectedWaveforms`, `Cases`, `Elements`, `Methods`, `SecondaryUnits` and `ElsIfBranches`.

pyVHDLModel/Sequential.py
Paebbels: Same fix - `SelectedExpressions[0]` / `SelectedExpressions[1]`, plus the selector is now marked as `Expression` (the actual field) rather than "the selector".

pyVHDLModel/Sequential.py
Paebbels: Done, using your 2D suggestion - see the reply on `ElseBranch` for the full rendering.

pyVHDLModel/Sequential.py
Paebbels: Done - same treatment as `IfBranch`.

pyVHDLModel/Sequential.py
Paebbels: Adopted your 2D suggestion for all three branch classes. The full if statement is shown, a bracket marks the part the class represents, and the individual fields are marked inside it:

```
      if sel = '0' then     -- ┐ IfBranch
      -- ^^^^^^^^^          -- │   <- Condition
        s <= '0';           -- │
      --^^^^^^^^^           -- ┘   <- Statements
      elsif sel = '1' then
        s <= '1';
      else
        s <= '0';
      end if;
```

`ElsifBranch` and `ElseBranch` bracket their own rows of the same statement. The bracket column is itself inside a VHDL comment, so the snippet still analyzes - I checked the rendered output with `ghdl -a --std=08` rather than assuming it.

pyVHDLModel/Sequential.py
Paebbels: Good question. I went with `<- Condition (optional)` - trailing, so the field name stays first and the marker column still reads as a list of field names. `optional Condition` buries the name behind an adjective.

Applied to `WaitStatement` (`Condition`, `Timeout`), `SequentialAssertStatement` (`Message`, `Severity`) and the statement labels. Easy to flip if you prefer the other order.

pyVHDLModel/Sequential.py
Paebbels: Renamed. `clk` appeared in three places (`ConcurrentProcedureCall`, `ConcurrentBlockStatement`, `WaitStatement`) and all three now use `clock`.

pyVHDLModel/Expression.py
Paebbels: Agreed. Dropped the type prefixes across the whole file - `boperand`/`voperand`/`soperand` are all `operand` now, and the result is `res` everywhere instead of `bres`/`vres`/`sres`. 45 classes were affected; the caret runs were regenerated so they follow the shorter names.

pyVHDLModel/Expression.py
Paebbels: Done - `res := integer(val);`.

pyVHDLModel/Expression.py
Paebbels: Fixed - `res := lhs ** rhs;`. I applied the same to the shift/rotate operators, which had a literal `1` as the right operand for the same reason (`res := lhs srl rhs;` etc.).

pyVHDLModel/Expression.py
Paebbels: Done - part of the same sweep, `res := lhs and rhs;`.

pyVHDLModel/Expression.py
Paebbels: Done - `res := lhs ?/= rhs;`.

Paebbels:
Second review round addressed in 77e5235; all 18 comments have replies.

Two of your comments turned out to be instances of a wider problem rather than one-offs, so I searched for the pattern instead of fixing the spot:

* The misplaced `^` (`s`/`v`) was a **shift that was not applied when the target sits in column 0** - it hit **six** examples, not two. My caret checker had passed them because it only verified a caret sits on a non-space character, and `<` qualifies. It now also rejects caret runs made purely of operator characters.
* The `b`/`v`/`s` prefixes ran through **45 classes** in `Expression.py`. All regenerated, carets re-computed for the shorter names.

I also added a round-trip check: every example block is re-rendered from its own extracted markers and compared against what is in the file. That found nine blocks in `Type.py` whose arrows had drifted out of alignment or that named fields in a trailing `-- Comment` instead of a marker. All are regenerated; the check now reports 0.

One example was not just mismarked but **not analyzable**: `AnonymousType` showed `type ptr;` on its own, and an incomplete type must be completed in the same declarative part. It now shows the completion.

Validation:

| Check | Result |
|---|---|
| Caret alignment / operator-only carets | 0 / 0 |
| Example blocks that do not round-trip | 0 |
| `seealso` links | 494 links / 180 blocks, 0 dangling, 0 duplicated, 0 mixed-format |
| `interrogate` | 70.1% (PASSED, minimum 59.0%) |
| GHDL `-a --std=08` on every changed construct | clean |

Two open questions in the threads: whether the optional **label** should be shown on all statement examples or only representative ones, and whether `X (optional)` or `optional X` is the marker wording you want.

Findings written up for the follow-up PR: the **block port map** gap you spotted (with the `# TODO: extract to mixin` in `Instantiation` as the fix path), and the **documentation coverage** breakdown - 361 interrogate misses categorized by effort, plus 204 undocumented class fields that interrogate does not count at all.

Paebbels:
Both answers applied in 791fcc1.

**1) Label shown in all statement examples.** That came to **34** classes, not 25 - the sweep found 31 without a label plus the 3 that already had one. One distinction I made rather than treating them uniformly: for `block`, the three generate statements and the three instantiation statements the label is **mandatory**, so those are marked `<- Label`; everything else gets `<- optional Label`. I checked which is which with the analyzer instead of assuming.

**2) `optional X` wording.** Applied to all 41 optional markers - `optional Label`, `optional Condition`, `optional Severity`, `optional Message`, `optional Timeout`, `optional SensitivityList`, `optional ReturnValue`, `optional LoopReference`, `optional Architecture`, and the `optional Subtype` in the alias example in `Declaration.py`. No `(optional)` remains.

Two things the sweep turned up that you had not asked about:

* The **if-generate and case-generate examples were not analyzable**: I had written the branch conditions and the case selector as signals, but a generate condition and a generate selector must be *static* expressions. Both now use constants (`WIDTH > 8`, `case MODE`), and the prose says so.
* The **concurrent** assignment examples still carried the old "first alternative"/"final branch" wording - I had only fixed the sequential ones in the previous round. They now use the list-element convention too.

Validation:

| Check | Result |
|---|---|
| Caret alignment / operator-only carets | 0 / 0 |
| Example blocks that do not round-trip | 0 |
| ReST parse | 0 problems |
| `interrogate` | 70.1% (PASSED) |
| GHDL `-a --std=08` on all 34 labelled statements | clean |

PR #152, opened 2026-07-27
Document every class field with a `#:` comment

Paebbels:
First of the follow-up items from the findings file after #151.

## What

All **204 undocumented class fields** now carry a `#:` comment, taking the package from **78/282 (27.7%)** to **282/282 (100%)**.

`Regions.py` and `__init__.py` were already complete and served as the style model: a `#:` comment on the same line, aligned per class.

## Why this is a separate workstream from the interrogate number

Worth being explicit, because it looks like the coverage should have moved: **interrogate does not count class fields at all.** Its 1207 items are modules, classes, functions/methods and properties only. So this PR leaves the number at 70.1% while closing a gap that is entirely invisible to it. The interrogate-visible work (361 misses, dominated by 246 `__init__` doc-strings) is the next PR.

The findings file recommends this order deliberately: most `__init__` parameters map 1:1 onto a field, so having the field docs first makes the `__init__` `:param:` pass substantially mechanical instead of 246 independent acts of authorship.

## Preserved TODO/FIXME notes

Five fields already carried a `# TODO:`/`# FIXME:` comment on the same line. Rather than drop the note or lose the documentation, the note moved to its own line above the field:

* `TernaryExpression._FORMAT`
* `ConcurrentStatementsMixin._instantiations`
* `ProcessStatement._sensitivityList`
* `ProcedureCallMixin._procedure`
* `Name._root`

## Verification

This change is **comment-only**, and that is checked rather than asserted: for every module, the Python token stream with comments removed is byte-identical to the one on `dev`.

| Check | Result |
|---|---|
| Modules whose code (comments stripped) changed | **0** |
| Class fields documented | 282/282 (100%) |
| Unbalanced ``` `` ``` or unterminated roles in `#:` comments | 0 |
| `interrogate` | 70.1% (unchanged, as expected) |

### DIFF
--- a/pyVHDLModel/Expression.py
+++ b/pyVHDLModel/Expression.py
@@ -121,7 +121,7 @@ class EnumerationLiteral(Literal):
 	      st <= Idle;
 	      --    ^^^^    <- Value
 	"""
-	_value: str
+	_value: str  #: The enumeration literal's name.
 
 	def __init__(self, value: str, parent: Nullable[ModelEntity] = None) -> None:
 		super().__init__(parent)
@@ -170,7 +170,7 @@ class IntegerLiteral(NumericLiteral):
 	      res := a + 42;
 	      --         ^^    <- Value
 	"""
-	_value: int
+	_value: int  #: The literal's integer value.
 
 	def __init__(self, value: int) -> None:
 		super().__init__()
@@ -203,7 +203,7 @@ class FloatingPointLiteral(NumericLiteral):
 	      r <= 3.14;
 	      --   ^^^^    <- Value
 	"""
-	_value: float
+	_value: float  #: The literal's floating-point value.
 
 	def __init__(self, value: float) -> None:
 		super().__init__()
@@ -242,7 +242,7 @@ class PhysicalLiteral(NumericLiteral):
 	   * :class:`Physical integer literal <pyVHDLModel.Expression.PhysicalIntegerLiteral>`
 	   * :class:`Physical floating literal <pyVHDLModel.Expression.PhysicalFloatingLiteral>`
 	"""
-	_unitName: str
+	_unitName: str  #: The name of the physical unit the value is given in.
 
 	def __init__(self, unitName: str) -> None:
 		super().__init__()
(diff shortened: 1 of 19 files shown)


pyVHDLModel/Expression.py
```diff
@@ -427,11 +427,11 @@ class BitStringLiteral(Literal):
...
-	_value:       str
-	_binaryValue: str
-	_bits:        int
-	_length:      Nullable[int]
-	_signed:      Nullable[bool]
+	_value:       str             #: The literal as written in the source, without the enclosing double quotes.
+	_binaryValue: str             #: The literal's value expanded to base 2, one character per bit.
+	_bits:        int             #: The number of bits the literal represents.
+	_length:      Nullable[int]   #: The explicitly given length, or ``None`` if the literal has no length specification.
+	_signed:      Nullable[bool]  #: ``True`` if signed, ``False`` if unsigned, ``None`` if unspecified.
```
Paebbels, line 434: for our todos, if not yet listed: the field and the property should be renamed to `_isSigned`/`IsSigned` to match the `Is***` -> bool format.

Paebbels: Added to the findings file as its own item, so it is not lost when this PR merges.

I surveyed every boolean-typed field and boolean-returning property to see how widespread it is. `_signed`/`Signed` is the **only** member stating a property without a predicate prefix - `IsPure`, `IsResolved`, `IsVHDL`, `IsAMS` and `HasPrefix` already follow the format. I left `AllowBlackbox` out of the rename: it reads as a permission flag rather than a predicate, so `IsAllowBlackbox` would be worse. Say the word if you want it included anyway.

The same sweep turned up a second, unrelated inconsistency: **`Component._isBlackBox` is the only place in the package spelling it `BlackBox`.** Everything else - including its own property `IsBlackbox`, plus `AllowBlackboxMixin`, `BlackboxWarning` - spells it `Blackbox`, so the field and its own property disagree. It appears at `DesignUnit.py:816`, `:837`, `:864`, `:896` and `__init__.py:1903`. I put it in the same findings entry since it is the same kind of fix.

I also checked the pyGHDL impact, and it is smaller than I expected: pyGHDL never reads the `Signed` property, and it constructs the bit-string literals **positionally** (`BinaryBitStringLiteral(literalNode, value, length, signed)` in `pyGHDL/dom/Literal.py:202`), so renaming the `__init__` parameter does not break it either. `_isBlackBox` does not appear in pyGHDL at all. `Signed` is still a public property, so the rename is breaking for other consumers and belongs in a release-noted naming PR rather than being slipped into an unrelated change.

PR #153, opened 2026-07-27
Document all 246 `__init__` methods, exclude nested functions from interrogate

Paebbels:
Workstream A from the findings file, continuing #152.

**Interrogate coverage: 70.1% -> 91.1%.**

## Nested functions excluded

As agreed, `ignore-nested-functions = true`. The eight remaining nested helpers (`Design.CreateTypeAndObjectGraph._LinkItems`, `Design.LinkComponents.linkStatements`, `Design.ComputeCompileOrder.predicate`, ...) are implementation-internal and never rendered by Sphinx. This alone took the item count from 1207 to 1199 and coverage to 70.6%.

## The 246 `__init__` doc-strings

Per the coding conventions, the class doc-string describes the class and `__init__` documents the construction parameters.

The `:param:` descriptions are **derived from the `#:` field documentation added in #152** rather than written independently: **642 of 678 parameters initialize a documented field**, so the field comment and the parameter description stay consistent by construction instead of by discipline. That is exactly the dependency the findings file predicted, and it is why the field pass went first.

The other 36 are parameters whose name differs from the field they feed (`rng` -> `_range`, `entitySymbol` -> `_entity`, `genericPackage` -> `_packageReference`, ...) or that have no backing field at all; those were written by hand.

Where an inherited description was accurate but useless in context I overrode it - e.g. `SignalForceAssignment.expression` said "The expression held by this construct" and now says "The value forced onto the signal"; the three selected-assignment classes now say "The selector expression".

## Pre-existing bugs found and fixed

The generator skips already-documented methods, so I checked those separately by comparing every `:param:` list against the actual signature. Three were wrong on `dev`:

* **`InterfaceGroup.__init__`** was titled *"Initialize a PortGroup with a list of ports and optional name"* - the wrong class - and documented **none** of its three parameters.
* **`BaseType.__init__`** was missing `:param documentation:`.
* **`Design.__init__`** listed its two parameters in the **wrong order**.

Also fixed a pre-existing typo, **"langauge" -> "language"**, in `Symbol.py`. It sat in a field comment that the generator copies, so it had already spread into 24 of the new doc-strings before I caught it (25 occurrences fixed).

## `fail-under` raised 59 -> 90

So the gain cannot silently regress. Flagging it explicitly since it was not part of the request - revert that one line if you would rather keep the gate low.

## Verification

| Check | Result |
|---|---|
| Modules whose AST (doc-strings stripped) changed | **0** |
| `:param:` lists disagreeing with the signature | **0** (was 3) |
| Example blocks: caret alignment / round-trip | 0 / 0 |
| `interrogate` | **91.1%** (PASSED, minimum 90.0%) |

## What is left

107 items: 60 `__str__`, 15 other dunders (`__repr__`, `__len__`, `__iter__`, `__bool__`), 32 real methods. All low-effort except the ~10 `Design.*` graph algorithms.

### DIFF
--- a/pyVHDLModel/Base.py
+++ b/pyVHDLModel/Base.py
@@ -681,7 +681,7 @@ class ChoicesMixin(metaclass=ExtendedType, mixin=True):
 
 	def __init__(self, choices: Nullable[Iterable[BaseChoice]] = None) -> None:
 		"""
-		Initializes a choices.
+		Initializes choices.
 
 		:param choices: List of all choices selecting this alternative.
 		"""
--- a/pyVHDLModel/Base.py
+++ b/pyVHDLModel/Base.py
@@ -487,6 +487,9 @@ class BranchMixin(metaclass=ExtendedType, mixin=True):
 	"""
 
 	def __init__(self) -> None:
+		"""
+		Initializes a branch.
+		"""
 		pass
 
 
@@ -501,6 +504,11 @@ class ConditionalBranchMixin(BranchMixin, ConditionalMixin, mixin=True):
 	   * :class:`Elsif branch mixin <pyVHDLModel.Base.ElsifBranchMixin>`
 	"""
 	def __init__(self, condition: ExpressionUnion) -> None:
+		"""
+		Initializes a conditional branch.
+
+		:param condition: The condition guarding this statement.
+		"""
 		super().__init__()
 		ConditionalMixin.__init__(self, condition)
 
@@ -556,6 +564,12 @@ class ReportStatementMixin(metaclass=ExtendedType, mixin=True):
 	_severity: Nullable[ExpressionUnion]  #: The reported severity level, or ``None`` if none was given.
 
 	def __init__(self, message: Nullable[ExpressionUnion] = None, severity: Nullable[ExpressionUnion] = None) -> None:
+		"""
+		Initializes a report statement.
+
+		:param message:  The reported message, or ``None`` if none was given.
+		:param severity: The reported severity level, or ``None`` if none was given.
+		"""
 		self._message = message
 		if message is not None:
 			message.Parent = self
@@ -595,6 +609,13 @@ class AssertStatementMixin(ReportStatementMixin, ConditionalMixin, mixin=True):
 	"""
 
 	def __init__(self, condition: ExpressionUnion, message: Nullable[ExpressionUnion] = None, severity: Nullable[ExpressionUnion] = None) -> None:
+		"""
+		Initializes an assert statement.
+
+		:param condition: The condition guarding this statement.
+		:param message:   The reported message, or ``None`` if none was given.
+		:param severity:  The reported severity level, or ``None`` if none was given.
+		"""
 		super().__init__(message, severity)
 		ConditionalMixin.__init__(self, condition)
 
@@ -609,6 +630,9 @@ class BlockStatementMixin(metaclass=ExtendedType, mixin=True):
 	"""
 
 	def __init__(self) -> None:
+		"""
+		Initializes a block statement.
+		"""
 		pass
 
 
@@ -656,6 +680,11 @@ class ChoicesMixin(metaclass=ExtendedType, mixin=True):
 	_choices: List[BaseChoice]  #: List of all choices selecting this alternative.
 
 	def __init__(self, choices: Nullable[Iterable[BaseChoice]] = None) -> None:
+		"""
+		Initializes a choices.
+
+		:param choices: List of all choices selecting this alternative.
+		"""
 		self._choices = []
 		if choices is not None:
 			for choice in choices:
@@ -820,6 +849,13 @@ class WaveformElement(ModelEntity):
 	_after: ExpressionUnion       #: The delay after which the value is assigned, or ``None`` if none was given.
 
 	def __init__(self, expression: ExpressionUnion, after: Nullable[ExpressionUnion] = None, parent: Nullable[ModelEntity] = None) -> None:
+		"""
+		Initializes a waveform element.
+
+		:param expression: The value this waveform element assigns.
+		:param after:      The delay after which the value is assigned, or ``None`` if none was given.
+		:param parent:     The parent model entity of this entity.
+		"""
 		super().__init__(parent)
 
 		self._expression = expression
--- a/pyVHDLModel/Common.py
+++ b/pyVHDLModel/Common.py
@@ -78,6 +78,11 @@ class AllowBlackboxMixin(metaclass=ExtendedType, mixin=True):
 	_allowBlackbox: Nullable[bool]  #: Allow blackboxes for components in language entity.
 
 	def __init__(self, allowBlackbox: Nullable[bool] = None) -> None:
+		"""
+		Initializes an allow blackbox.
+
+		:param allowBlackbox: Allow blackboxes for components in language entity.
+		"""
 		self._allowBlackbox = allowBlackbox
 
 	@property
@@ -120,6 +125,12 @@ class Statement(ModelEntity, LabeledEntityMixin):
 	   * :class:`Sequential statement <pyVHDLModel.Sequential.SequentialStatement>`
 	"""
 	def __init__(self, label: Nullable[str] = None, parent=None) -> None:
+		"""
+		Initializes a statement.
+
+		:param label:  The label of a model entity.
+		:param parent: The parent model entity of this entity.
+		"""
 		super().__init__(parent)
 		LabeledEntityMixin.__init__(self, label)
 
@@ -142,6 +153,12 @@ class ProcedureCallMixin(metaclass=ExtendedType, mixin=True):
 	_parameterAssociationItems: List[ParameterAssociationItem]  #: List of all parameter associations of the call.
 
 	def __init__(self, procedureName: Symbol, parameterAssociationItems: Nullable[Iterable[ParameterAssociationItem]] = None) -> None:
+		"""
+		Initializes a procedure call.
+
+		:param procedureName:             Reference to the called procedure.
+		:param parameterAssociationItems: List of all parameter associations of the call.
+		"""
 		self._procedure = procedureName
 		procedureName.Parent = self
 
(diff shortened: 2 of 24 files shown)


pyVHDLModel/__init__.py
```diff
@@ -2650,6 +2650,15 @@ def __init__(
 		library:       Nullable[Library] = None,
 		parent:        Nullable[ModelEntity] = None
 	) -> None:
+		"""
+		Initializes a document.
+
+		:param path:          path to the document. ``None`` if virtual document.
```
Paebbels:
```suggestion
		:param path:          Path to the document. ``None`` if in-memory document.
```
Start description with upper-case letter.

pyVHDLModel/__init__.py
```diff
@@ -2650,6 +2650,15 @@ def __init__(
 		library:       Nullable[Library] = None,
 		parent:        Nullable[ModelEntity] = None
 	) -> None:
+		"""
+		Initializes a document.
```
Paebbels:
```suggestion
		Initializes a VHDL document.
```

pyVHDLModel/Base.py
```diff
@@ -656,6 +680,11 @@ class ChoicesMixin(metaclass=ExtendedType, mixin=True):
 	_choices: List[BaseChoice]  #: List of all choices selecting this alternative.
 
 	def __init__(self, choices: Nullable[Iterable[BaseChoice]] = None) -> None:
+		"""
+		Initializes a choices.
```
Paebbels: a -> singular vs. choices -> plural

```suggestion
		Initializes choices.
```

pyVHDLModel/Common.py
```diff
@@ -78,6 +78,11 @@ class AllowBlackboxMixin(metaclass=ExtendedType, mixin=True):
 	_allowBlackbox: Nullable[bool]  #: Allow blackboxes for components in language entity.
 
 	def __init__(self, allowBlackbox: Nullable[bool] = None) -> None:
+		"""
+		Initializes an allow blackbox.
```
Paebbels: Strange grammar and meaning.

```suggestion
		Initializes a hierarchical model entity allow for blackboxes.
```

pyVHDLModel/IEEE.py
```diff
@@ -103,6 +103,12 @@ class Ieee(PredefinedLibrary):
 	_flavor: IEEEFlavor  #: The flavor of the ``ieee`` library this instance provides.
 
 	def __init__(self, flavor: Nullable[IEEEFlavor] = None) -> None:
+		"""
+		Initializes an ieee.
```
Paebbels:
```suggestion
		Initializes the ``ieee`` library.
```

There is only one predefined library called like this. Similar for library Std.

pyVHDLModel/IEEE.py
```diff
@@ -169,6 +175,9 @@ class Math_Complex(PredefinedPackage):
 	"""
 
 	def __init__(self) -> None:
+		"""
+		Initializes a math complex.
```
Paebbels:
```suggestion
		Initializes the ``math_complex`` package.
```

There is only one predefined package called like this. There are multiple application in this file and in Std.py

pyVHDLModel/Object.py
```diff
@@ -71,6 +71,14 @@ class Obj(ModelEntity, MultipleNamedEntityMixin, DocumentedEntityMixin):
 	_objectVertex: Nullable[Vertex]  #: The vertex representing this object in the design's object graph.
 
 	def __init__(self, identifiers: Iterable[str], subtype: Symbol, documentation: Nullable[str] = None, parent: Nullable[ModelEntity] = None) -> None:
+		"""
+		Initializes an obj.
```
Paebbels:
```suggestion
		Initializes an object.
```

Needed to name it Obj to avoid collisions with Python's object.

pyVHDLModel/STD.py
```diff
@@ -59,6 +59,9 @@ class Std(PredefinedLibrary):
 	"""
 
 	def __init__(self) -> None:
+		"""
+		Initializes a std.
```
Paebbels:
```suggestion
		Initializes the ``std`` library.
```

pyVHDLModel/STD.py
```diff
@@ -84,6 +87,9 @@ class Standard(PredefinedPackage):
 	"""
 
 	def __init__(self) -> None:
+		"""
+		Initializes a standard.
```
Paebbels:
```suggestion
		Initializes the ``standard`` package.
```

pyVHDLModel/STD.py
```diff
@@ -84,6 +87,9 @@ class Standard(PredefinedPackage):
...
+		"""
 		super().__init__()
 
 		boolean = EnumeratedType("boolean", (EnumerationLiteral("false"), EnumerationLiteral("true")), None)
```
Paebbels, line 95: Add to todos: translate VHDL-2008 packages into preinitialized items. currently only a handfull of types and sub programs are defined in these packages.

pyVHDLModel/__init__.py
Paebbels: Applied. The lower-case start came from the field comment it is derived from (`_path` on line 2628), so I fixed both - otherwise the next regeneration would reintroduce it. I checked the whole package for the same fault: no other `:param:`, `:returns:` or `:raises:` description starts lower-case.

pyVHDLModel/__init__.py
Paebbels: Applied - "Initializes a VHDL document."

pyVHDLModel/Base.py
Paebbels: Applied. I searched for the same fault rather than fixing just this one: **nine** mixins are named after a plural and all read "Initializes a <plural>". They now read "Initializes choices", "Initializes concurrent statements", "Initializes sequential statements", "Initializes conditional waveforms", "Initializes selected waveforms" and "Initializes selected expressions". The three `With*Mixin` classes were the same shape ("Initializes a with generics") and now say "Initializes a language construct with a generic clause", matching their class doc-strings.

pyVHDLModel/Common.py
Paebbels: Applied your wording verbatim.

pyVHDLModel/IEEE.py
Paebbels: Applied, and generalized. Rather than hand-editing the ones you spotted, I now derive these summaries from **each class doc-string**, which already states the VHDL identifier (`Predefined VHDL library ``ieee``.`). That regenerated **30** summaries across `IEEE.py` and `STD.py` - every predefined library, package and package body, not just the reported ones.

One detail worth noting: I first lower-cased the identifiers, which would have produced ``vital_timing``. The class doc-strings spell them ``VITAL_Timing``, and that is how IEEE 1076.4 names them, so the generator now preserves the doc-string casing.

pyVHDLModel/IEEE.py
Paebbels: Covered by the same change - all package classes now read "Initializes the ``<name>`` package", and the body classes "Initializes the ``<name>`` package body", taking the name from the class doc-string.

pyVHDLModel/Object.py
Paebbels: Applied - "Initializes an object." Thanks for the context on the name; that is exactly the case a class-name-derived summary cannot get right.

pyVHDLModel/STD.py
Paebbels: Applied - "Initializes the ``std`` library."

pyVHDLModel/STD.py
Paebbels: Applied - "Initializes the ``standard`` package."

pyVHDLModel/STD.py
Paebbels: Added to the findings file as its own item.

I captured the scope you described - `Standard` currently sets up only `boolean`, `bit`, `integer`, `character`, `severity_level`, `time`, `string` and a few vector types, so everything else a real `std.standard` declares (operator overloads, conversion functions, and the equivalent in `numeric_std`, `math_real`, ...) is absent and name resolution against a predefined package finds nothing for most identifiers.

I noted one decision as needing to come first, because it dominates the effort: hand-write the declarations, or **generate** them by parsing GHDL's own `std`/`ieee` sources with pyGHDL and emitting the model construction code. Hand-writing is accurate but large and will drift from the LRM; generating costs more up front but stays correct and can be regenerated per revision. Also suggested splitting it per package so `std.standard` can land on its own.

Paebbels:
All 10 review comments addressed in 7092d70.

They were symptoms of one cause: I derived each summary by **splitting the class name into words**. The class name is simply the wrong source, so the fix was to stop using it rather than to patch the reported cases.

* **Predefined libraries and packages** now take their name from the class doc-string, which already states the VHDL identifier (`Predefined VHDL library ``ieee``.`). That regenerated **30** summaries across `IEEE.py` and `STD.py`, not the three reported.
* **Everywhere else**, the summary is derived from the class doc-string's `Represents a ...` line when it has one, so the initializer and the class agree. That corrected **98** more, including hyphenation (`if generate` -> `if-generate`) and word order (`concurrent simple signal assignment` -> `simple concurrent signal assignment`).
* **Plural-named mixins**: nine read "Initializes a <plural>"; all fixed.
* `Obj`, `Document`, `WithDefaultExpressionMixin`, `AllowBlackboxMixin` per your suggestions.

Three things worth flagging:

* I first lower-cased the predefined identifiers, which would have produced ``vital_timing``. The class doc-strings spell them ``VITAL_Timing``, which is how **IEEE 1076.4** names them, so the generator preserves the doc-string casing.
* `UnaryExpression` read "an unary expression". My article rule keyed off the first letter being a vowel, but "unary" starts with a consonant *sound*. Fixed, and I checked every summary for both directions of article mismatch - 0 remain.
* The `path` description started lower-case because the **field comment** it derives from does. I fixed both, otherwise a regeneration would reintroduce it. No other `:param:`/`:returns:`/`:raises:` description in the package starts lower-case.

**A pre-existing bug this surfaced**: deriving `Component`'s summary from its class doc-string produced *"Initializes a configuration declaration"* - because `Component`'s **class** doc-string says "Represents a configuration declaration". That is a copy-paste from `Configuration` that I introduced in #151 and which is already merged to `dev`. Corrected here to "a component declaration".

Validation unchanged: 0 modules whose AST (doc-strings stripped) differs from `dev`, 0 param/signature mismatches, ReST clean, examples intact (carets and round-trip both 0), 0 added lines over 120 chars, 428 tests passing, interrogate 91.1%.

The VHDL-2008 predefined-package request is recorded in the findings file for its own PR.

PR #154, opened 2026-07-27
Document all 75 dunder methods

Paebbels:
Continues #153. **Interrogate coverage: 91.1% -> 97.3%.**

Documented 60 `__str__`, 8 `__repr__`, 3 `__len__`, 3 `__iter__` and 1 `__bool__`, following the style the already-documented dunders use: summary, a `**Format:**` line showing the rendered output, and `:returns:`.

The **32 regular methods are deliberately left alone** as agreed - they will be touched again when the algorithms and their tests are extended, and documenting them now means doing it twice.

## The `**Format:**` examples come from running the code

I constructed real model objects and captured the actual output rather than describing what the code looked like it did. That was worth doing - **three of the formats I first wrote from reading the code were wrong**. For example the interface groups actually render:

```
GenericGroup myGroup (2) - generics: WIDTH, DEPTH)
```

not the shorter form the source suggests at a glance.

## Four defects surfaced; one fixed here

**Fixed in this PR**, because documenting it otherwise meant describing a method that raises:

* **`Component.__repr__` returned `None`** - and therefore raised `TypeError: __repr__ returned non-string` - whenever the parent was neither a `Package` nor an `Architecture`, e.g. for a freshly constructed component. `repr(Component("counter"))` crashed. Both branches of the `isinstance` chain were **identical**, so the fix is to drop the chain entirely.

**Recorded in the findings file** instead, because each is a rendering or modelling decision rather than an obvious correction:

* `Context` and `PackageBody` render `mylib?.myContext`. The `?` is appended when the parent **is** present - inverted relative to the `%`-for-absent-parent convention `Package`, `Entity`, `Architecture` and `Configuration` use. Looks like a copy-paste inversion, but which convention wins is your call.
* The three interface groups emit an **unbalanced closing parenthesis**, and disagree among themselves about the colon (`PortGroup:` vs `GenericGroup`).
* **`BitStringLiteral.__str__` raises `AttributeError` on the base class**: `_base` is a `ClassVar` on the four concrete subclasses only. Its `BitStringBase.NoBase` branch is unreachable today, which is fairly clear evidence the base class was meant to carry `_base = NoBase`. Held back because it decides whether that base class is instantiable at all.

Say the word and I will fix all three in a follow-up.

## Verification

| Check | Result |
|---|---|
| Modules whose AST (doc-strings stripped) changed | **1** - `DesignUnit.py`, the `Component.__repr__` fix |
| Dunders returning a value without `:returns:` | 0 |
| `interrogate` | **97.3%** (PASSED, minimum 90.0%) |

I left `fail-under` at 90 rather than raising it to 97, since the remaining 32 methods will move the number again shortly.

### DIFF
--- a/pyVHDLModel/Name.py
+++ b/pyVHDLModel/Name.py
@@ -135,9 +135,23 @@ class Name(ModelEntity):
 		return self._prefix is not None
 
 	def __repr__(self) -> str:
+		"""
+		Formats a representation of the name.
+
+		**Format:** ``Name: 'sig'``
+
+		:returns: String representation of the name.
+		"""
 		return f"Name: '{self.__str__()}'"
 
 	def __str__(self) -> str:
+		"""
+		Formats the name.
+
+		**Format:** ``sig``
+
+		:returns: Formatted name.
+		"""
 		return self._identifier
 
 
@@ -186,6 +200,13 @@ class ParenthesisName(Name):
 		return self._associations
 
 	def __str__(self) -> str:
+		"""
+		Formats the name.
+
+		**Format:** ``func(a, b)``
+
+		:returns: Formatted name.
+		"""
 		return f"{self._prefix!s}({', '.join(str(a) for a in self._associations)})"
 
 
@@ -228,6 +249,13 @@ class IndexedName(Name):
 		return self._indices
 
 	def __str__(self) -> str:
+		"""
+		Formats the name.
+
+		**Format:** ``arr(0)``
+
+		:returns: Formatted name.
+		"""
 		return f"{self._prefix!s}({', '.join(str(i) for i in self._indices)})"
 
 
@@ -271,6 +299,13 @@ class SelectedName(Name):
 		super().__init__(identifier, prefix, parent)
 
 	def __str__(self) -> str:
+		"""
+		Formats the name.
+
+		**Format:** ``rec.elem``
+
+		:returns: Formatted name.
+		"""
 		return f"{self._prefix!s}.{self._identifier}"
 
 
@@ -297,6 +332,13 @@ class AttributeName(Name):
 		super().__init__(identifier, prefix, parent)
 
 	def __str__(self) -> str:
+		"""
+		Formats the name.
+
+		**Format:** ``v'range``
+
+		:returns: Formatted name.
+		"""
 		return f"{self._prefix!s}'{self._identifier}"
 
 
@@ -333,4 +375,11 @@ class OpenName(Name):
 		super().__init__("open", parent=parent)  # TODO: the case of 'OPEN' is not preserved
 
 	def __str__(self) -> str:
+		"""
+		Formats the name.
+
+		**Format:** ``open``
+
+		:returns: Formatted name.
+		"""
 		return "open"
--- a/pyVHDLModel/Name.py
+++ b/pyVHDLModel/Name.py
@@ -201,11 +201,11 @@ class ParenthesisName(Name):
 
 	def __str__(self) -> str:
 		"""
-		Formats the name.
+		Formats the parenthesis name.
 
 		**Format:** ``func(a, b)``
 
-		:returns: Formatted name.
+		:returns: Formatted parenthesis name.
 		"""
 		return f"{self._prefix!s}({', '.join(str(a) for a in self._associations)})"
 
@@ -250,11 +250,11 @@ class IndexedName(Name):
 
 	def __str__(self) -> str:
 		"""
-		Formats the name.
+		Formats the indexed name.
 
 		**Format:** ``arr(0)``
 
-		:returns: Formatted name.
+		:returns: Formatted indexed name.
 		"""
 		return f"{self._prefix!s}({', '.join(str(i) for i in self._indices)})"
 
@@ -300,11 +300,11 @@ class SelectedName(Name):
 
 	def __str__(self) -> str:
 		"""
-		Formats the name.
+		Formats the selected name.
 
 		**Format:** ``rec.elem``
 
-		:returns: Formatted name.
+		:returns: Formatted selected name.
 		"""
 		return f"{self._prefix!s}.{self._identifier}"
 
@@ -333,11 +333,11 @@ class AttributeName(Name):
 
 	def __str__(self) -> str:
 		"""
-		Formats the name.
+		Formats the attribute name.
 
 		**Format:** ``v'range``
 
-		:returns: Formatted name.
+		:returns: Formatted attribute name.
 		"""
 		return f"{self._prefix!s}'{self._identifier}"
 
@@ -376,10 +376,10 @@ class OpenName(Name):
 
 	def __str__(self) -> str:
 		"""
-		Formats the name.
+		Formats the open name.
 
 		**Format:** ``open``
 
-		:returns: Formatted name.
+		:returns: Formatted open name.
 		"""
 		return "open"
--- a/pyVHDLModel/Symbol.py
+++ b/pyVHDLModel/Symbol.py
@@ -47,7 +47,7 @@ from pyVHDLModel.Name      import Name, AllName
 @export
 class PossibleReference(Flag):
 	"""
-	Is an enumeration, representing possible targets for a reference in a :mod:`pyVHDLModel.Symbol`.
+	Is an enumeration, representing possible targets for a reference in a :class:`~pyVHDLModel.Symbol.Symbol`.
 	"""
 
 	Unknown =         0
--- a/pyVHDLModel/Symbol.py
+++ b/pyVHDLModel/Symbol.py
@@ -47,7 +47,7 @@ from pyVHDLModel.Name      import Name, AllName
 @export
 class PossibleReference(Flag):
 	"""
-	Is an enumeration, representing possible targets for a reference in a :class:`~pyVHDLModel.Symbol`.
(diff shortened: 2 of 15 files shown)


Paebbels:
On (1) - I checked this rather than agreeing, and **I cannot confirm it**. The convention you describe is real, but it is not what produces the `?` here.

**The unresolved-symbol marker does exist**, exactly as you say: `Symbol.__str__` returns `f"{self._name!s}?"` while `_reference is None`, which is why an unlinked symbol prints `clk?`.

**But no `Symbol` is on this code path.** A design unit's `_parent` is a `Library` *object*, and `_identifier` is a plain `str`. The `?` is a literal appended in the f-string expression:

```py
# Context.__str__ and PackageBody.__str__
lib = self._parent._identifier + "?" if self._parent is not None else ""
```

The ternary binds as `(self._parent._identifier + "?") if (self._parent is not None) else ""` - so the `?` is added precisely when the parent **is** present.

Demonstrated by construction:

| Case | `Context` | `Package` |
|------|-----------|-----------|
| attached to a fully resolved library | `Context: mylib?.myContext` | `Package: 'mylib.myPackage'` |
| `_parent is None` | `Context: .myContext` | `Package: '%.myPackage'` |

In row 1 nothing whatsoever is unresolved - the library is a real, attached `Library` - and it still prints `mylib?`. In row 2, where the parent genuinely **is** missing, the marker vanishes entirely, while `Package` correctly shows `%`.

So the logic is inverted relative to what the marker convention would mean, and it is inconsistent with the four design units that use `%`-for-absent (`Package`, `Entity`, `Architecture`, `Configuration`).

Your point that `mylib?` and `mylib` are similar enough for testing still holds - nothing is blocked by this. I have updated the findings entry with the evidence above so the follow-up starts from the right diagnosis rather than from "it is the symbol marker".

Noted on (2) and (3) going to the follow-up, and on keeping `fail-under` at 90. This PR is unchanged and ready.

Paebbels:
Fixed in 7f46bab - you were right that the precedence was wrong, and I have made the convention uniform.

**`Context` / `PackageBody`** now behave as you specified:

| | before | after |
|---|---|---|
| library set | `Context: mylib?.myContext` | `Context: mylib.myContext` |
| no parent | `Context: .myContext` | `Context: ?.myContext` |

**`%` -> `?`** everywhere else. That was **12 more sites** than just `Package`, because the same placeholder appears in `Entity`, `Architecture` and `Configuration` too - and not only for the library. I applied it to all of them so one convention holds:

```
Entity: 'mylib.myEntity(?)'      <- entity with no known architecture
Architecture: ?.myEntity(rtl)   <- architecture whose library is unknown
```

If you meant `?` only for the *parent* and wanted the architecture list to keep its own marker, say so and I will split them again - but a single marker seemed closer to your intent.

The convention is now stated once on the `DesignUnit` base-class rather than left implicit, and the `**Format:**` examples were updated.

**One test was asserting the old output**: `Instantiation/Model.py::Symbols::test_EntityInstantiationSymbol` expected `Entity: 'liB.enT(%)'`, now `(?)`. Worth noting my first check for affected tests missed it - I grepped for `"%"` and the literal is `(%)` inside a longer string. The suite caught it; my grep did not. It now reads consistently with the assertion directly above it, which already expected the symbol's own `Lib.Ent?`.

**Also repaired 16 dangling Sphinx references**, all pre-existing on `dev`, found by a link check over the new doc-strings:

* `pyVHDLModel.DesignUnits.*` -> `pyVHDLModel.DesignUnit.*` (8 - stale module name)
* `Std_logic_arith` / `_misc` / `_signed` / `_textio` / `_unsigned` -> `Std_Logic_*` (6 - casing)
* `Type.SubType` -> `Type.Subtype` (1)
* `:class:`~pyVHDLModel.Symbol`` -> `:mod:`pyVHDLModel.Symbol`` (1 - it names a module, not a class)

That last fix uncovered something for the follow-up list, now in the findings file: **`IEEE.py` defines `Std_Logic_Arith` and `Std_Logic_TextIO` twice each** (Mentor Graphics at lines 512/257, Synopsys at 656/712). Python keeps the last binding, so the first of each pair is unreachable by name and `@export` registers only one - which is why those references had nowhere unambiguous to point.

| Check | Result |
|---|---|
| Dangling `:class:` references | **0** (was 16) |
| ReST parse | 0 problems |

pyVHDLModel/Name.py
```diff
@@ -271,6 +299,13 @@ def __init__(self, identifier: str, prefix: Name, parent: Nullable[ModelEntity]
 		super().__init__(identifier, prefix, parent)
 
 	def __str__(self) -> str:
+		"""
+		Formats the name.
```
Paebbels:
```suggestion
		Formats the selected name.
```

pyVHDLModel/Name.py
```diff
@@ -297,6 +332,13 @@ def __init__(self, identifier: str, prefix: Name, parent: Nullable[ModelEntity]
...
```
Paebbels:
```suggestion
		Formats the attribute name.
```

pyVHDLModel/Symbol.py
```diff
@@ -47,7 +47,7 @@
 @export
 class PossibleReference(Flag):
 	"""
-	Is an enumeration, representing possible targets for a reference in a :class:`~pyVHDLModel.Symbol`.
+	Is an enumeration, representing possible targets for a reference in a :mod:`pyVHDLModel.Symbol`.
```
Paebbels: I think I wanted to reference the class Symbol within module Symbol, because this uses PossibleReference, right?

pyVHDLModel/Type.py
```diff
@@ -600,6 +637,13 @@ def ElementType(self) -> Symbol:
 		return self._elementType
 
 	def __str__(self) -> str:
+		"""
+		Formats the type definition.
```
Paebbels:
```suggestion
		Formats the array type definition.
```

pyVHDLModel/Type.py
```diff
@@ -651,6 +695,13 @@ def Subtype(self) -> Symbol:
 		return self._subtype
 
 	def __str__(self) -> str:
+		"""
+		Formats the element declaration.
```
Paebbels:
```suggestion
		Formats the record element declaration.
```

pyVHDLModel/Name.py
Paebbels: Applied - and the same fault ran through the whole file. I had written these summaries per *file group* rather than per class, so every name in `Name.py` said "Formats the name". All five are now specific: parenthesis, indexed, selected, attribute and open name.

pyVHDLModel/Name.py
Paebbels: Covered by the same change.

pyVHDLModel/Type.py
Paebbels: Applied, and likewise for the rest of `Type.py` - all ten said "Formats the type definition". Now enumerated / integer / floating-point / physical / array / record / access / file type definition, plus the record element declaration and the subtype declaration.

pyVHDLModel/Type.py
Paebbels: Applied - "Formats the record element declaration."

pyVHDLModel/Symbol.py
Paebbels: You are right, and my change was the wrong reading - reverted to a class reference, now fully qualified as `:class:`~pyVHDLModel.Symbol.Symbol`` so it resolves.

For context on why I touched it: the original was `:class:`~pyVHDLModel.Symbol``, which points at the *module* path with a `:class:` role, so it was one of 16 references that resolved to nothing. I read it as "the module was meant" and switched the role; the right fix was to complete the path instead. `PossibleReference` is exactly what the `Symbol` class uses to filter its references, as you say.

Paebbels:
All five comments addressed in 1361481.

The four "be more specific" ones were the same fault repeated: I had written the summaries **per file group** rather than per class, so entire groups shared one noun - every name in `Name.py` said *"Formats the name"*, every type in `Type.py` said *"Formats the type definition"*. Fixing only the four you marked would have left the rest inconsistent, so I made **all 36** specific:

| File | Count | Example |
|---|---|---|
| `Name.py` | 5 | "Formats the selected name." |
| `Type.py` | 10 | "Formats the array type definition." |
| `Expression.py` | 10 | "Formats the ternary expression." |
| `Concurrent.py` | 4 | "Formats the ranged case-generate choice." |
| `Sequential.py` | 4 | "Formats the ``others`` case alternative." |
| `Base.py` | 3 | "Formats the range denoted by a name." |

The 36 matching `:returns:` lines carried the same generic nouns and are synced to their summaries. I added a check that every `__str__`/`__repr__` summary shares at least one word with its class name - it now reports 0 violations.

**Found in that sweep**: `Mode.__str__` said *"Formats the direction"* and returned *"Formatted direction"* - copy-pasted from `Direction.__str__`, pre-existing on `dev`. Fixed to "mode".

On `PossibleReference`: you were right and my change was the wrong reading, now reverted to a class reference and fully qualified as `:class:`~pyVHDLModel.Symbol.Symbol`` so it actually resolves. Worth explaining why I touched it at all - the original was `:class:`~pyVHDLModel.Symbol``, a `:class:` role pointing at a *module* path, so it was one of the 16 references resolving to nothing. I read that as "the module was meant"; the right fix was to complete the path instead.

| Check | Result |
|---|---|
| Summaries not naming their class | **0** |
| Dangling `:class:` references | 0 |
| ReST parse | 0 problems |

PR #155, opened 2026-07-27
Fix two `__str__` defects, unify boolean naming, de-duplicate IEEE class names

Paebbels:
Findings items **3, 5 and 6**. Item **7** turned out to be already fixed - see the bottom.

⚠️ **Needs the companion pyGHDL PR** (linked below) - the `Signed` rename breaks pyGHDL's testsuite.

## Item 3 - two `__str__` defects

* The three interface groups emitted an **unbalanced closing parenthesis**, and disagreed on the colon. All three now render `<Group>: <name> (<n>) - <kind>: <names>`.
* **`BitStringLiteral.__str__` raised `AttributeError` on the base class** - `_base` was a `ClassVar` on the four subclasses only. Declared on the base as `BitStringBase.NoBase`, which also makes the method's existing `NoBase` branch reachable. A commented-out `# _base:  ClassVar[BitStringBase]` sat above the class doc-string, confirming the intent; removed.

## Item 5 - boolean naming

* `_signed`/`Signed` -> `_isSigned`/`IsSigned`, including the `__init__` parameter. `BitStringBase.Signed` is an enum member and untouched.
* `Component._isBlackBox` -> `_isBlackbox` - the only `BlackBox` spelling in the package, disagreeing with its own property `IsBlackbox`.

## Item 6 - the naive rename would have corrupted the VHDL names

This is the part worth reading. `Std_Logic_Arith` and `Std_Logic_TextIO` were each defined twice, so the first of each pair was unreachable and `@export` registered only one.

But **the Python class name *is* the VHDL identifier**: `PredefinedPackage.__init__` passes `self.__class__.__name__`, and `PredefinedPackageBody.__init__` uses `self.__class__.__name__[:-5]`. Renaming alone produced:

```
std_logic_arith_synopsys              <- wrong package name
std_logic_arith_body_mentorgra        <- truncated by [:-5]
```

So the identifier is **decoupled first**: both base classes now take an optional explicit identifier, defaulting to the previous class-name derivation. Only then are the four classes renamed with `_MentorGraphics`/`_Synopsys` suffixes, passing their VHDL name explicitly. `Std_Logic_Arith_Body_MentorGraphics` gained the `__init__` it needed.

**Verified behaviour-neutral**: the VHDL identifiers registered for every flavor (standard, Synopsys, Mentor Graphics, VITAL) are byte-identical to `dev`.

Worth noting the runtime was never broken - each `*_PACKAGES` tuple is built *before* the next definition shadows the name, so the right classes were captured. The collision only affected module-level names, `__all__` and Sphinx.

## Item 7 - already fixed, findings entry was stale

The three `@readonly` getters that define a setter were changed to `@property` in **79ce362** (PR #149) - the same PR whose findings entry says the decorators were "deliberately not changed". A check across the package reports **0** remaining, so there was nothing to do.

## Verification

| Check | Result |
|---|---|
| Registered VHDL identifiers vs `dev` | **identical** for all four flavors |
| `@readonly` getters defining a setter | 0 |
| Duplicate class names in `IEEE.py` | 0 (was 2) |
| ReST parse | 0 problems |

Tests updated for the intended renames and the corrected group format (9 assertions).

### DIFF
--- a/pyVHDLModel/Interface.py
+++ b/pyVHDLModel/Interface.py
@@ -1042,12 +1042,12 @@ class GenericGroup(InterfaceGroup, WithGenericsMixin):
 		"""
 		Formats the generic group.
 
-		**Format:** ``GenericGroup myGroup (2) - generics: WIDTH, DEPTH)``
+		**Format:** ``GenericGroup: myGroup (2) - generics: WIDTH, DEPTH``
 
 		:returns: Formatted generic group.
 		"""
 		names = ", ".join(name for item in self._genericItems for name in identifiersOf(item))
-		return f"GenericGroup {self._identifier} ({len(self._genericItems)}) - generics: {names})"
+		return f"GenericGroup: {self._identifier} ({len(self._genericItems)}) - generics: {names}"
 
 
 @export
@@ -1100,12 +1100,12 @@ class PortGroup(InterfaceGroup, WithPortsMixin):
 		"""
 		Formats the port group.
 
-		**Format:** ``PortGroup: myGroup (2) - ports: clock, reset)``
+		**Format:** ``PortGroup: myGroup (2) - ports: clock, reset``
 
 		:returns: Formatted port group.
 		"""
 		names = ", ".join(name for item in self._portItems for name in identifiersOf(item))
-		return f"PortGroup: {self._identifier} ({len(self._portItems)}) - ports: {names})"
+		return f"PortGroup: {self._identifier} ({len(self._portItems)}) - ports: {names}"
 
 
 @export
@@ -1158,9 +1158,9 @@ class ParameterGroup(InterfaceGroup, WithParametersMixin):
 		"""
 		Formats the parameter group.
 
-		**Format:** ``ParameterGroup myGroup (2) - parameters: a, b)``
+		**Format:** ``ParameterGroup: myGroup (2) - parameters: a, b``
 
 		:returns: Formatted parameter group.
 		"""
 		names = ", ".join(name for item in self._parameterItems for name in identifiersOf(item))
-		return f"ParameterGroup {self._identifier} ({len(self._parameterItems)}) - parameters: {names})"
+		return f"ParameterGroup: {self._identifier} ({len(self._parameterItems)}) - parameters: {names}"
--- a/pyVHDLModel/Interface.py
+++ b/pyVHDLModel/Interface.py
@@ -1042,12 +1042,12 @@ class GenericGroup(InterfaceGroup, WithGenericsMixin):
 		"""
 		Formats the generic group.
 
-		**Format:** ``GenericGroup: myGroup (2) - generics: WIDTH, DEPTH``
+		**Format:** ``GenericGroup: myGroup (2): WIDTH, DEPTH``
 
 		:returns: Formatted generic group.
 		"""
 		names = ", ".join(name for item in self._genericItems for name in identifiersOf(item))
-		return f"GenericGroup: {self._identifier} ({len(self._genericItems)}) - generics: {names}"
+		return f"GenericGroup: {self._identifier} ({len(self._genericItems)}): {names}"
 
 
 @export
@@ -1100,12 +1100,12 @@ class PortGroup(InterfaceGroup, WithPortsMixin):
 		"""
 		Formats the port group.
 
-		**Format:** ``PortGroup: myGroup (2) - ports: clock, reset``
+		**Format:** ``PortGroup: myGroup (2): clock, reset``
 
 		:returns: Formatted port group.
 		"""
 		names = ", ".join(name for item in self._portItems for name in identifiersOf(item))
-		return f"PortGroup: {self._identifier} ({len(self._portItems)}) - ports: {names}"
+		return f"PortGroup: {self._identifier} ({len(self._portItems)}): {names}"
 
 
 @export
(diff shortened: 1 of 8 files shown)


pyVHDLModel/Interface.py
```diff
@@ -1042,12 +1042,12 @@ def __str__(self) -> str:
 		"""
 		Formats the generic group.
 
-		**Format:** ``GenericGroup myGroup (2) - generics: WIDTH, DEPTH)``
+		**Format:** ``GenericGroup: myGroup (2) - generics: WIDTH, DEPTH``
```
```suggestion
		**Format:** ``GenericGroup: myGroup (2): WIDTH, DEPTH``
```

pyVHDLModel/Interface.py
```diff
@@ -1042,12 +1042,12 @@ def __str__(self) -> str:
...
 
 		:returns: Formatted generic group.
 		"""
 		names = ", ".join(name for item in self._genericItems for name in identifiersOf(item))
-		return f"GenericGroup {self._identifier} ({len(self._genericItems)}) - generics: {names})"
+		return f"GenericGroup: {self._identifier} ({len(self._genericItems)}) - generics: {names}"
```
```suggestion
		return f"GenericGroup: {self._identifier} ({len(self._genericItems)}): {names}"
```

pyVHDLModel/Interface.py
```diff
@@ -1100,12 +1100,12 @@ def __str__(self) -> str:
 		"""
 		Formats the port group.
 
-		**Format:** ``PortGroup: myGroup (2) - ports: clock, reset)``
+		**Format:** ``PortGroup: myGroup (2) - ports: clock, reset``
```
```suggestion
		**Format:** ``PortGroup: myGroup (2): clock, reset``
```

pyVHDLModel/Interface.py
```diff
@@ -1100,12 +1100,12 @@ def __str__(self) -> str:
...
 
 		:returns: Formatted port group.
 		"""
 		names = ", ".join(name for item in self._portItems for name in identifiersOf(item))
-		return f"PortGroup: {self._identifier} ({len(self._portItems)}) - ports: {names})"
+		return f"PortGroup: {self._identifier} ({len(self._portItems)}) - ports: {names}"
```
```suggestion
		return f"PortGroup: {self._identifier} ({len(self._portItems)}): {names}"
```

pyVHDLModel/Interface.py
```diff
@@ -1158,9 +1158,9 @@ def __str__(self) -> str:
 		"""
 		Formats the parameter group.
 
-		**Format:** ``ParameterGroup myGroup (2) - parameters: a, b)``
+		**Format:** ``ParameterGroup: myGroup (2) - parameters: a, b``
```
```suggestion
		**Format:** ``ParameterGroup: myGroup (2): a, b``
```

pyVHDLModel/Interface.py
```diff
@@ -1158,9 +1158,9 @@ def __str__(self) -> str:
...
 
 		:returns: Formatted parameter group.
 		"""
 		names = ", ".join(name for item in self._parameterItems for name in identifiersOf(item))
-		return f"ParameterGroup {self._identifier} ({len(self._parameterItems)}) - parameters: {names})"
+		return f"ParameterGroup: {self._identifier} ({len(self._parameterItems)}) - parameters: {names}"
```
```suggestion
		return f"ParameterGroup: {self._identifier} ({len(self._parameterItems)}): {names}"
```

pyVHDLModel/Predefined.py
```diff
@@ -119,11 +119,16 @@ class PredefinedPackage(Package, PredefinedPackageMixin):
...
 		"""
 		Initializes a predefined package.
+
+		By default the VHDL package name is the Python class name. Pass ``identifier`` when the two must
+		differ - e.g. when two vendor flavors of the same VHDL package need distinct Python classes.
+
+		:param identifier: The VHDL package name, or ``None`` to use the class name.
 		"""
-		super().__init__(self.__class__.__name__, parent=None)
+		super().__init__(self.__class__.__name__ if identifier is None else identifier, parent=None)
```
Paebbels:
```suggestion
		super().__init__(self.__class__.__name__ if identifier is None else identifier)
```

pyVHDLModel/Interface.py
Paebbels: Applied.

pyVHDLModel/Interface.py
Paebbels: Applied - the class name already carries the item kind, so repeating it was redundant.

pyVHDLModel/Interface.py
Paebbels: Applied.

pyVHDLModel/Interface.py
Paebbels: Applied.

pyVHDLModel/Interface.py
Paebbels: Applied.

pyVHDLModel/Interface.py
Paebbels: Applied. All three groups now render `<Group>: <name> (<n>): <names>`, and the six test expectations follow.

pyVHDLModel/Predefined.py
Paebbels: Applied. I did the same in `PredefinedPackageBody.__init__`, which passed `parent=None` to `PackageBody.__init__` for the same reason - your suggestion named only the package, but both parameters already default to `None`. Say the word if you would rather I left the body alone.

Paebbels:
All seven comments applied in 9dca48a.

The group formats are now:

```
GenericGroup: myGroup (2): WIDTH, DEPTH
PortGroup: myGroup (2): clock, reset
ParameterGroup: myGroup (2): a, b
```

Repeating the item kind after the class name was redundant - and it was my addition in #154 that made it worse by adding the colon while leaving `- generics:` in place. Six test expectations updated to match.

`parent=None` dropped in **both** `PredefinedPackage` and `PredefinedPackageBody`; your suggestion named only the package, but the body passed it to `PackageBody.__init__` for the same reason, and both parameters already default to `None`.

| Check | Result |
|---|---|
| Registered VHDL identifiers vs `dev` | still **identical** for all four flavors |
| ReST parse | 0 problems |

Companion Paebbels/GHDL#16 is merged, so #155 is unblocked.

PR #156, opened 2026-07-28
Give the naming mixins a `__str__`, let the interface groups delegate

Paebbels:
The findings item queued for its own branch, now that the group `__str__` methods have settled.

## What

`NamedEntityMixin` and `MultipleNamedEntityMixin` render themselves - the identifier, and the comma-joined identifiers respectively.

**56 classes gained a `__str__`.** Every one of them previously fell back to Python's default `<pyVHDLModel.X.Y object at 0x...>`:

```
Signal                        <pyVHDLModel.Object.Signal object at 0x...>   ->  s1, s2
Procedure                     <pyVHDLModel.Subprogram.Procedure object ...>  ->  myProcedure
PortSimpleSignalInterfaceItem <pyVHDLModel.Interface.Port... object at ...>  ->  p
```

Classes with a more specific `__str__` (`Entity`, `Architecture`, `Subtype`, `DeferredConstant`, `RecordTypeElement`, the predefined packages, ...) keep priority via the MRO and are untouched.

## The payoff

`GenericGroup`/`PortGroup`/`ParameterGroup.__str__` now simply join `str(item)` instead of reaching into identifier internals via `identifiersOf()`:

```py
names = ", ".join(name for item in self._genericItems for name in identifiersOf(item))
names = ", ".join(str(item) for item in self._genericItems)
```

**Rendered output is unchanged** - `identifiersOf` flattened an item's identifiers into separate names, and `MultipleNamedEntityMixin.__str__` joins them with the same separator, so a two-identifier item still contributes `G2, G3`.

`identifiersOf()` is kept - it is exported public API - but is no longer used inside the package.

## The risk this carried, and why it does not materialize

The finding flagged that adding `__str__` to two mixins inherited by dozens of classes could change something currently user-visible. One case looked exactly like that: `Library` renders `Library: 'mylib'` today, which is *not* a default object repr.

It turns out `Design`, `Library` and `Document` alias **`__str__ = __repr__`** explicitly, so the mixin never applies to them. My initial static scan missed this by looking only for `def __str__`; running the code caught it.

Verified properly: I recorded `str()` for an instance of every affected class before and after the change. **The only differences are default object reprs becoming names** - nothing that already rendered sensibly changed.

## Verification

| Check | Result |
|---|---|
| Classes whose `str()` changed | 56, all from `<object at 0x...>` to a name |
| Previously-sensible `str()` output that regressed | **0** |
| Group rendering vs before | identical |
| ReST parse | 0 problems |

### DIFF
--- a/pyVHDLModel/Base.py
+++ b/pyVHDLModel/Base.py
@@ -213,6 +213,19 @@ class NamedEntityMixin(metaclass=ExtendedType, mixin=True):
 		"""
 		return self._normalizedIdentifier
 
+	def __str__(self) -> str:
+		"""
+		Formats the named entity as its identifier.
+
+		Derived classes rendering more than their name - a full declaration, say - override this and take
+		priority via the MRO.
+
+		**Format:** ``myEntity``
+
+		:returns: The entity's identifier.
+		"""
+		return self._identifier
+
 
 @export
 class OptionallyNamedEntityMixin(metaclass=ExtendedType, mixin=True):
@@ -303,6 +316,19 @@ class MultipleNamedEntityMixin(metaclass=ExtendedType, mixin=True):
 		:returns: Tuple of normalized identifiers.
 		"""
 		return self._normalizedIdentifiers
+
+	def __str__(self) -> str:
+		"""
+		Formats the named entity as its identifiers.
+
+		A single declaration may name several entities at once, so all identifiers are rendered. Derived
+		classes rendering more than their names override this and take priority via the MRO.
+
+		**Format:** ``a, b``
+
+		:returns: The entity's identifiers, comma-separated.
+		"""
+		return ", ".join(self._identifiers)
 
 
 @export
--- a/pyVHDLModel/Interface.py
+++ b/pyVHDLModel/Interface.py
@@ -41,7 +41,7 @@ from pyTooling.MetaClasses  import ExtendedType
 
 from pyVHDLModel.Symbol     import Symbol, SubtypeSymbol, ModeViewSymbol
 from pyVHDLModel.Base       import ModelEntity, DocumentedEntityMixin, NamedEntityMixin, OptionallyNamedEntityMixin
-from pyVHDLModel.Base       import MultipleNamedEntityMixin, identifiersOf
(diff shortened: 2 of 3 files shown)


Paebbels:
I noticed classes or the primary inheritance tree is responsible to provide dunder-str. A mixin brings fields, properties and maybe methods (algorithms), but it's not responsible for the external view or representation of the class.

Paebbels:
Understood, and agreed - that is a cleaner separation than what I proposed. A mixin contributes state and behaviour; how the object presents itself belongs to the class's own hierarchy. I have recorded it as a standing convention so I do not reach for a mixin again.

I have rewritten the findings entry to mark the mixin approach as **rejected, do not retry**, quoting your reasoning, and to describe the shape that respects it.

Worth noting the underlying problem is bigger than the 56 classes this PR touched: **~173 classes have no `__str__` anywhere in their MRO**, so they render as `<pyVHDLModel.X.Y object at 0x...>`. They group under relatively few primary bases, so it is not 173 methods:

| Primary base | Classes below it | Note |
|---|---|---|
| `ModelEntity` | 30 | too generic to render usefully - handle subtrees instead |
| `SequentialStatement` / `ConcurrentStatement` / `Statement` | ~25 | rendering is the VHDL syntax; best done per statement |
| `VHDLModelException` / `Warning` | 14 | already carry a message - probably leave alone |
| `Flag` / `Enum` | 9 | `Direction`/`Mode`/`ObjectClass` already have one |
| **`Obj`** | 5 | one method covers `Constant`/`Signal`/`Variable`/`SharedVariable`/`File` |
| `Subprogram`, `BaseType`, `Constraint`, `BaseChoice`, `EntityAspect`, `ModeViewElement`, ... | 2-4 each | small and independent |

Doing it on the primary bases also allows something more useful than a bare name. `DeferredConstant` already shows the pattern - it renders `constant dc : bit`, not just `dc` - so `Obj.__str__` could render `<objectClass> <identifiers> : <subtype>` for all five object kinds in one place.

These are independent, so they need not be one PR. Unless you would rather I move straight on to the `ProtectedType`/`ProtectedTypeBody` item (next in the order you gave), I would start with `Obj` plus the interface items, since those are what the groups actually render.

PR #157, opened 2026-07-28
Give the primary base classes a `__str__`

Paebbels:
Replaces #156, which you closed because the external representation of a class belongs to its **primary inheritance tree**, not to a mixin that contributes fields, properties and algorithms. Recorded as a standing convention on my side.

## What

52 named classes rendered as `<pyVHDLModel.X.Y object at 0x...>`. They descend from **ten** primary bases, so ten methods cover all of them:

| Base | Covers |
|---|---|
| `Obj` | `Constant`, `Variable`, `Signal`, `SharedVariable`, `File` + every interface item derived from them |
| `BaseType` | the whole type tree, incl. `ProtectedType`/`ProtectedTypeBody` |
| `Subprogram` | procedures, functions, methods, their instantiations |
| `DesignUnit`, `Alias`, `Attribute`, `InterfacePackage`, `ModeViewDeclaration`, `ModeViewElement`, `DefaultClock` | the rest |

Classes with a more specific `__str__` keep priority via the MRO.

## Two judgement calls worth flagging

**`Obj` renders `s1, s2 : bit`, without an object-class keyword.** Rendering `constant`/`signal`/... would be more informative for declarations - and `DeferredConstant` already does exactly that - but interface items inherit from `Constant`/`Signal`/`Variable`/`File`, so a keyword would render a **port** as `signal p : bit` and state something the source does not. The chosen shape matches what `RecordTypeElement` already used. Showing the interface *mode* instead (`p : in bit`) is the better refinement and is recorded in the findings file.

**The interface groups keep `identifiersOf()`.** Letting them join `str(item)` was the payoff the original finding wanted, but that only works while items render as bare names. Now that they render a declaration, delegating would change group output from `p1, p2` to `p1 : bit, p2 : bit`. Group rendering is verified unchanged.

## Verification

I recorded `str()` for an instance of every affected class before and after.

| Check | Result |
|---|---|
| Classes whose `str()` changed | 52, all from `<object at 0x...>` to something meaningful |
| Previously-sensible output that regressed | **0** |
| Group rendering vs `dev` | identical |
| Mixins defining `__str__`/`__repr__` | **0** (asserted by a check) |
| ReST parse | 0 problems |

### DIFF
--- a/pyVHDLModel/Declaration.py
+++ b/pyVHDLModel/Declaration.py
@@ -131,13 +131,13 @@ class Attribute(ModelEntity, NamedEntityMixin, DocumentedEntityMixin):
 
 	def __str__(self) -> str:
 		"""
-		Formats the attribute as its identifier.
+		Formats the attribute declaration.
 
-		**Format:** ``myAttribute``
+		**Format:** ``attribute myAttribute: bit``
 
-		:returns: The attribute's identifier.
+		:returns: Formatted attribute declaration.
 		"""
-		return self._identifier
+		return f"attribute {self._identifier}: {self._subtype}"
 
 
 @export
@@ -309,10 +309,11 @@ class Alias(ModelEntity, NamedEntityMixin, DocumentedEntityMixin):
 
 	def __str__(self) -> str:
 		"""
-		Formats the alias as its identifier.
+		Formats the alias declaration.
 
-		**Format:** ``myAlias``
+		**Format:** ``alias myAlias : bit is target``, or without the subtype when none was given
 
-		:returns: The alias' identifier.
+		:returns: Formatted alias declaration.
 		"""
-		return self._identifier
+		subtype = f" : {self._subtype}" if self._subtype is not None else ""
+		return f"alias {self._identifier}{subtype} is {self._name}"
--- a/pyVHDLModel/Declaration.py
+++ b/pyVHDLModel/Declaration.py
@@ -129,6 +129,16 @@ class Attribute(ModelEntity, NamedEntityMixin, DocumentedEntityMixin):
 		"""
 		return self._subtype
 
+	def __str__(self) -> str:
+		"""
+		Formats the attribute as its identifier.
+
+		**Format:** ``myAttribute``
+
+		:returns: The attribute's identifier.
+		"""
+		return self._identifier
+
 
 @export
 class AttributeSpecification(ModelEntity, DocumentedEntityMixin):
@@ -296,3 +306,13 @@ class Alias(ModelEntity, NamedEntityMixin, DocumentedEntityMixin):
 		:returns: The subtype, or ``None`` if not set.
 		"""
 		return self._subtype
+
+	def __str__(self) -> str:
+		"""
+		Formats the alias as its identifier.
+
+		**Format:** ``myAlias``
+
+		:returns: The alias' identifier.
+		"""
+		return self._identifier
--- a/pyVHDLModel/DesignUnit.py
+++ b/pyVHDLModel/DesignUnit.py
@@ -37,7 +37,7 @@ Design units are contexts, entities, architectures, packages and their bodies as
 from typing import List, Dict, Union, Iterable, Optional as Nullable
 
 from pyTooling.Decorators   import export, readonly
-from pyTooling.MetaClasses  import ExtendedType
+from pyTooling.MetaClasses  import ExtendedType, abstractmethod
 from pyTooling.Graph        import Vertex
 
 from pyVHDLModel.Common     import AllowBlackboxMixin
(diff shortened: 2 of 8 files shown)


pyVHDLModel/Declaration.py
```diff
@@ -129,6 +129,16 @@ def Subtype(self) -> None:
...
 
+	def __str__(self) -> str:
+		"""
+		Formats the attribute as its identifier.
+
+		**Format:** ``myAttribute``
+
+		:returns: The attribute's identifier.
+		"""
+		return self._identifier
```
Paebbels:
```suggestion
		return f"attribute {self._identifier}: {self._subtype}"
```

I think a pure identifier isn't beneficial. We could emit the reconstructed statement, especially for statements or language constructs easily represented by a single line assuming normal coding styles.

Therefore an entity isn't printed with all it's generics, ports, declared items and statements, but here an attribute, an alias, a type, a subtype, or a constant could be recreated in dunder str.

pyVHDLModel/Interface.py
```diff
@@ -221,6 +231,16 @@ def Elements(self) -> List[ModeViewElement]:
...
 
+	def __str__(self) -> str:
+		"""
+		Formats the mode view declaration as its identifier.
+
+		**Format:** ``myView``
+
+		:returns: The mode view's identifier.
+		"""
+		return self._identifier
```
```suggestion
		elements = ", ".join(self....)
		return f"Mode view {self._identifier}: elements"
```

Here, we can't return the whole thing, but we can indicate what it is and list all element names.

pyVHDLModel/DesignUnit.py
```diff
@@ -350,6 +350,16 @@ def HierarchyVertex(self) -> Vertex:
...
 
+	def __str__(self) -> str:
+		"""
+		Formats the design unit as its identifier.
+
+		**Format:** ``myEntity``
+
+		:returns: The design unit's identifier.
+		"""
+		return self._identifier
```
Paebbels:
```suggestion
	@abstractmethod
	def __str__(self) -> str:
		pass
```
No `__str__` for design unit, because overwritten in all derived variants. You can even mark such a method, to ensure it's overwritten in all derived classes.

No need to specify a dunder method, if it's never used, because all instantiated derived classes will have their own dunder-str.

@abstractmethod is from pyTooling as well.

I'm not sure is pyTooling (I wrote it myself - but it's big) supports a `@abstractclass` or `@abstract` to mark the `DesignUnit` class as abstract independent of a method. (we had this problem at other places, where you identified classes as intermediate clases in the class hierarchy, but never being directly instantiated.

Paebbels: Applied - `@abstractmethod` from `pyTooling.MetaClasses`, so `ExtendedType` now rejects instantiating a unit that forgets it.

I checked what the base implementation actually covered before removing it, because "overwritten in all derived variants" is not obvious from the class list: of the **57** `DesignUnit` subclasses, only **three** resolved to it - `PrimaryUnit`, `SecondaryUnit` and `PSLPrimaryUnit`, all intermediates. The predefined packages and `PackageInstantiation` look like they lack a `__str__`, but they inherit `Package.__str__` through the MRO.

One consequence worth knowing: `tests/unit/Instantiation/DesignUnit.py` **did** instantiate `DesignUnit`, `PrimaryUnit` and `SecondaryUnit` directly - deliberately, to cover the base-class plumbing (context-item separation, the `Document`/`Library` setters). Six tests failed on `AbstractClassError`. They now go through minimal concrete stand-ins that add nothing but the required `__str__`, so the coverage is kept and nothing abstract is instantiated.

On your question about marking a whole class abstract: `pyTooling.MetaClasses` exports `abstractmethod` and `mustoverride` (plus `AbstractClassError`/`MustOverrideClassError`), but no `@abstractclass`/`@abstract` class decorator - a class becomes abstract by having an un-overridden `@abstractmethod`. The difference between the two: `abstractmethod` replaces the body with one raising `NotImplementedError`, while `mustoverride` keeps the original callable via `super()`, for a partial implementation. So for the intermediate classes you mentioned elsewhere, the only lever today is an abstract method.

pyVHDLModel/Declaration.py
Paebbels: Applied in 2c95310, along with the principle behind it - `__str__` now reconstructs a single-line declaration wherever the construct fits on one: `constant c1, c2 : bit`, `signal s : bit`, `alias myAlias : bit is target`, `type myType`, `package myPackage`, `procedure myProcedure(a, b)`, `default clock myClock`.

pyVHDLModel/Interface.py
Paebbels: Applied in 2c95310 as `view myView of myRecord: a, b` - it names itself and lists the element names, as you suggested. I used the VHDL-2019 spelling (`view ... of ...`) rather than "Mode view" so it reads like the construct it represents.

PR #159, opened 2026-07-28
Make protected types and their bodies declaration regions

Paebbels:
Findings item 3. Branched off `dev`, so it is independent of #157.

## What the analyzer says

Measured with `ghdl -a --std=19`, one declaration at a time, to settle whether either half can share an existing region rather than guessing:

| Declaration | protected type | protected type **body** | subprogram body |
|---|---|---|---|
| variable, constant | rejected | allowed | allowed |
| **signal, shared variable** | rejected | **rejected** | **rejected** |
| type, subtype, alias, file | rejected | allowed | allowed |
| nested subprogram *body* | rejected | allowed | allowed |
| subprogram *declaration* | **allowed** | rejected | rejected |
| attribute, use clause | rejected | allowed | allowed |

Your instinct was right, and it cuts cleanly both ways:

* A protected type **body** and a subprogram declarative part **agree on all twelve items**, so `ProtectedTypeBody` reuses `SequentialDeclarationRegionMixin`. Its indexed kinds - types, subtypes, constants, variables, files, functions, procedures, and notably *no* signals and *no* shared variables - match the measured set exactly.
* A protected type **declaration** is the exact complement: subprogram declarations and nothing else, the narrowest declarative region in VHDL. It gets a new `ProtectedTypeDeclarationRegionMixin`.

## The `_methods` conflation is resolved

`ProtectedTypeBody` stored **every** declared item in `_methods`, so a variable was indistinguishable from a method. Both classes now have a real `_declaredItems`, and `Methods` is its subprogram subset.

`ProtectedType.__init__`'s first list parameter is renamed `methods` -> `declaredItems` to match. pyGHDL passes it positionally and is unaffected (its suite is green).

## `Subprogram.Parent` no longer special-cases protected types

The isinstance guard carried a comment saying "a protected type is no declaration region, hence the check". That is no longer true. A method's namespace now chains to its enclosing region like any other subprogram, so **a method variable hides a body variable of the same name** - the `-Whide` case that motivated this finding in the first place. Verified directly.

## One thing worth reviewing carefully: the import cycle

`Type` must inherit the region mixins, but `Regions` **and** `Namespace` both imported `Type` at module level. I deferred their `Type` imports into the methods that need them - the same pattern `IndexDeclaredItems` already used for `Subprogram` - and quoted the affected annotations.

Quoting matters beyond style here: an unquoted `-> BaseType` or `Dict[str, FullType]` is evaluated **eagerly on Python 3.11-3.13** and would raise `NameError` there while passing on 3.14, exactly the CI failure pattern from #145. I checked by forcing evaluation of every annotation in the package: all clean.

## Verification

| Check | Result |
|---|---|
| Forced annotation evaluation (3.11-3.13 behaviour) | all clean |
| Method variable hides body variable | works |
| `pytest tests/unit` | **430** passed (+2), 69 subtests |
| ReST parse | 0 problems |
| New lines > 120 chars | 0 (the remaining long lines were already over on `dev`) |

The two protected-type tests passed a bare `ModelEntity` as a method stand-in, which `Methods` now correctly excludes. Rewritten with real `Procedure` instances and extended to cover the `DeclaredItems`/`Methods` distinction and namespace indexing.

### DIFF
--- a/pyVHDLModel/Regions.py
+++ b/pyVHDLModel/Regions.py
@@ -34,7 +34,7 @@ This module contains parts of an abstract document language model for VHDL.
 
 tbd.
 """
-from typing                 import List, Dict, Iterable, Optional as Nullable, Any
+from typing                 import TYPE_CHECKING, List, Dict, Iterable, Optional as Nullable, Any
 
 from pyTooling.Decorators   import export, readonly
 from pyTooling.MetaClasses  import ExtendedType
@@ -44,7 +44,8 @@ from pyVHDLModel.Base       import normalizedIdentifiersOf
 from pyVHDLModel.Exception  import NotImplementedWarning
 from pyVHDLModel.Namespace  import Namespace
 from pyVHDLModel.Object     import Constant, SharedVariable, File, Variable, Signal
-from pyVHDLModel.Type       import Subtype, FullType
+if TYPE_CHECKING:
+	from pyVHDLModel.Type     import Subtype, FullType
 
 # `pyVHDLModel.Subprogram` imports this module (Subprogram is a sequential declaration region), so
 # `Function`/`Procedure` are quoted in annotations and imported lazily where they're needed at runtime.
@@ -125,8 +126,8 @@ class ConcurrentDeclarationRegionMixin(DeclarationRegionMixin, mixin=True):
 
 	# _attributes:     Dict[str, Attribute]
 	# _aliases:        Dict[str, Alias]
-	_types:           Dict[str, FullType]               #: Dictionary of all types declared in this concurrent declaration region.
-	_subtypes:        Dict[str, Subtype]                #: Dictionary of all subtypes declared in this concurrent declaration region.
+	_types:           Dict[str, 'FullType']               #: Dictionary of all types declared in this concurrent declaration region.
+	_subtypes:        Dict[str, 'Subtype']                #: Dictionary of all subtypes declared in this concurrent declaration region.
 	# _objects:        Dict[str, Union[Constant, Variable, Signal]]
 	_constants:       Dict[str, Constant]               #: Dictionary of all constants declared in this concurrent declaration region.
 	_signals:         Dict[str, Signal]                 #: Dictionary of all signals declared in this concurrent declaration region.
@@ -173,7 +174,7 @@ class ConcurrentDeclarationRegionMixin(DeclarationRegionMixin, mixin=True):
 		return self._declaredItems
 
 	@readonly
-	def Types(self) -> Dict[str, FullType]:
+	def Types(self) -> Dict[str, 'FullType']:
 		"""
 		Read-only property to access the types (:attr:`_types`).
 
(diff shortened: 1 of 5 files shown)


pyVHDLModel/Regions.py
```diff
@@ -125,8 +126,8 @@ class ConcurrentDeclarationRegionMixin(DeclarationRegionMixin, mixin=True):
 
 	# _attributes:     Dict[str, Attribute]
 	# _aliases:        Dict[str, Alias]
-	_types:           Dict[str, FullType]               #: Dictionary of all types declared in this concurrent declaration region.
-	_subtypes:        Dict[str, Subtype]                #: Dictionary of all subtypes declared in this concurrent declaration region.
+	_types:           Dict[str, 'FullType']               #: Dictionary of all types declared in this concurrent declaration region.
```
Paebbels: Alignment got broken.

pyVHDLModel/Regions.py
```diff
@@ -364,8 +366,8 @@ class SequentialDeclarationRegionMixin(DeclarationRegionMixin, mixin=True):
 	_declaredItems: List                          #: List of all declared items in this sequential declaration region.
 	_namespace:     Namespace                     #: The namespace of this sequential declaration region.
 
-	_types:         Dict[str, FullType]           #: Dictionary of all types declared in this sequential declaration region.
-	_subtypes:      Dict[str, Subtype]            #: Dictionary of all subtypes declared in this sequential declaration region.
+	_types:         Dict[str, 'FullType']           #: Dictionary of all types declared in this sequential declaration region.
```
Paebbels: Alignment got broken.

pyVHDLModel/Regions.py
```diff
@@ -522,3 +525,107 @@ def IndexDeclaredItems(self) -> None:
...
+	.. note::
+
+	   VHDL's ``protected_type_declarative_item`` rule admits **subprogram declarations only** - it is the
+	   narrowest declarative region in the language. Verified with the analyzer: a protected type
+	   declaration rejects variables, constants, signals, shared variables, types, subtypes, aliases,
+	   files, attributes, use clauses and even a subprogram *body*, while a subprogram *declaration* is
+	   accepted. That is the exact complement of a protected type **body**, which shares the sequential
+	   region (:class:`SequentialDeclarationRegionMixin`) with subprogram bodies.
+
+	.. seealso::
```
Paebbels, line 539:
```suggestion

	.. note::

	.. seealso::
```

To chatty. Research result don't need to documented here. We assume, we both know VHDL well and write correct and and define a good architecture. Showing our research in official documentation outlines our missing knowledge about the language -> not good to present ourselfes like that.

pyVHDLModel/Regions.py
Paebbels: Fixed - quoting the annotation added two characters and pushed the `#:` column right. Padding adjusted in both blocks. Note `_functions`/`_procedures` in this same block sit two columns further right again, but they already did on `dev`, so I left them.

pyVHDLModel/Regions.py
Paebbels: Fixed in the same commit.

pyVHDLModel/Regions.py
Paebbels: Agreed, and cut. The doc-string now states the rule and the contrast with a protected type body, without recounting what was tried.

I applied the same to `ProtectedTypeBody`, which was listing the legal declarations for exactly the same reason. The measurements stay in the findings file, where they explain *why* the mixin is reused - useful to us, not to a reader of the API documentation.

PR #160, opened 2026-07-28
Detect duplicate declarations within a declarative region

Paebbels:
Findings item 1 - the last of the four you ordered, and the only genuine correctness gap of the set.

The model silently overwrote a namespace entry when two declarations shared an identifier. `Namespace.AddElement` now reports a `DuplicateDeclarationWarning`, and every insertion in the three region mixins goes through it.

## Which pairs are one region - measured, not assumed

| Construct | GHDL `-a --std=08` |
|---|---|
| two declarations in one declarative part | **error** |
| architecture declaration vs entity declaration | **error** |
| architecture declaration vs entity port | **error** |
| **package body declaration vs package declaration** | **error** |
| block / generate / process / subprogram | hiding (`-Whide`), legal |

The **package/body** pair was the open question in the finding. It is one region, so `_sharesRegionWithParent` marks that namespace link and the entity/architecture one. The check walks outwards only while that flag is set, so genuinely nested regions still hide as before, and resolution is unchanged - the innermost declaration still wins.

## Two traps worth reading

**Overloading.** Naive detection would flag legal VHDL. Measured:

| Case | GHDL |
|---|---|
| two procedures, different signature | legal |
| two procedures, identical signature | **error** |
| function and procedure, same name | legal |
| same enumeration literal in two types | legal |
| enumeration literal vs a signal | **error** |

Subprogram insertions pass `overloadable=True`. Signatures are **not** compared yet, so an identical-signature redeclaration is still missed - that needs real overload resolution, and is recorded in the findings file along with the enumeration-literal cases (literals are never indexed into a namespace today).

**Re-indexing.** `IndexDeclaredItems` is not idempotent, and `Analyze()` plus the explicit `Index*` entry points call it more than once. My first version reported **every predefined `std.standard` declaration as a duplicate**. Re-inserting the *same* object is now recognised as a re-index rather than a second declaration, and there is a test for it.

## Severity

`DuplicateDeclarationWarning` extends `VHDLModelCriticalWarning` - "likely indicates a defect in the model or its input", which fits. Reporting goes through `WarningCollector`, so it is visible to a caller that collects warnings and **silently dropped otherwise**, exactly like the existing warnings. Detection is opt-in and breaks no existing caller. Say the word if you would rather it raised.

## Verification

| Check | Result |
|---|---|
| `pytest tests/unit` | **442** passed (+12), 69 subtests |
| pyGHDL `testsuite/pyunit` (real VHDL, incl. the predefined libraries) | 185 passed, 17 xfailed - **no false positives** |
| ReST parse | 0 problems |
| New lines > 120 chars | 0 |

`tests/unit/Namespace/Hiding.py` previously documented this as a known gap; it now asserts the duplicate is reported. New `tests/unit/Namespace/Duplicates.py` covers the region pairs, overloading and re-indexing.

### DIFF
--- a/pyVHDLModel/DesignUnit.py
+++ b/pyVHDLModel/DesignUnit.py
@@ -34,7 +34,7 @@ This module contains parts of an abstract document language model for VHDL.
 
 Design units are contexts, entities, architectures, packages and their bodies as well as configurations.
 """
-from typing import List, Dict, Union, Iterable, Optional as Nullable
+from typing import ClassVar, List, Dict, Union, Iterable, Optional as Nullable
 
 from pyTooling.Decorators   import export, readonly
 from pyTooling.MetaClasses  import ExtendedType, abstractmethod
@@ -182,6 +182,8 @@ class DesignUnit(ModelEntity, NamedEntityMixin, DocumentedEntityMixin):
 	     * :class:`Package body <pyVHDLModel.DesignUnit.PackageBody>`
 	"""
 
+	_continuesParentRegion: ClassVar[bool] = False  #: ``True`` if it continues its parent's declarative region.
+
 	_document: 'Document'                                  #: The VHDL library, the design unit was analyzed into.
 
 	# Either written as statements before (e.g. entity, architecture, package, ...), or as statements inside (context)
@@ -236,7 +238,7 @@ class DesignUnit(ModelEntity, NamedEntityMixin, DocumentedEntityMixin):
 		self._dependencyVertex = None
 		self._hierarchyVertex = None
 
-		self._namespace = Namespace(self._normalizedIdentifier)
+		self._namespace = Namespace(self._normalizedIdentifier, sharesRegion=self._continuesParentRegion)
 
 	@property
 	def Document(self) -> 'Document':
@@ -650,6 +652,8 @@ class PackageBody(SecondaryUnit, DesignUnitWithContextMixin, ConcurrentDeclarati
 	   * :class:`Package it implements <pyVHDLModel.DesignUnit.Package>`
 	"""
 
+	_continuesParentRegion: ClassVar[bool] = True
+
 	_package:       PackageSymbol  #: Reference to the package this body implements.
 
 	def __init__(
@@ -844,6 +848,8 @@ class Architecture(SecondaryUnit, DesignUnitWithContextMixin, ConcurrentDeclarat
 
 	   * :class:`Entity it implements <pyVHDLModel.DesignUnit.Entity>`
 	"""
+
+	_continuesParentRegion: ClassVar[bool] = True
 
 	_entity:        EntitySymbol  #: Reference to the entity this architecture implements.
 
--- a/pyVHDLModel/DesignUnit.py
+++ b/pyVHDLModel/DesignUnit.py
@@ -182,7 +182,7 @@ class DesignUnit(ModelEntity, NamedEntityMixin, DocumentedEntityMixin):
 	     * :class:`Package body <pyVHDLModel.DesignUnit.PackageBody>`
 	"""
 
-	_continuesParentRegion: ClassVar[bool] = False  #: ``True`` if it continues its parent's declarative region.
+	_continuesParentRegion: ClassVar[bool] = False         #: ``True`` if it continues its parent's declarative region.
 
 	_document: 'Document'                                  #: The VHDL library, the design unit was analyzed into.
 
@@ -238,7 +238,7 @@ class DesignUnit(ModelEntity, NamedEntityMixin, DocumentedEntityMixin):
 		self._dependencyVertex = None
 		self._hierarchyVertex = None
 
-		self._namespace = Namespace(self._normalizedIdentifier, sharesRegion=self._continuesParentRegion)
+		self._namespace = Namespace(self._normalizedIdentifier, sharesRegionWithParent=self._continuesParentRegion)
 
 	@property
 	def Document(self) -> 'Document':
@@ -652,9 +652,9 @@ class PackageBody(SecondaryUnit, DesignUnitWithContextMixin, ConcurrentDeclarati
 	   * :class:`Package it implements <pyVHDLModel.DesignUnit.Package>`
 	"""
 
-	_continuesParentRegion: ClassVar[bool] = True
+	_continuesParentRegion: ClassVar[bool] = True   #: A package body continues its package's declarative region.
 
-	_package:       PackageSymbol  #: Reference to the package this body implements.
+	_package:               PackageSymbol           #: Reference to the package this body implements.
 
 	def __init__(
 		self,
@@ -849,9 +849,9 @@ class Architecture(SecondaryUnit, DesignUnitWithContextMixin, ConcurrentDeclarat
 	   * :class:`Entity it implements <pyVHDLModel.DesignUnit.Entity>`
 	"""
 
-	_continuesParentRegion: ClassVar[bool] = True
+	_continuesParentRegion: ClassVar[bool] = True  #: An architecture continues its entity's declarative region.
 
-	_entity:        EntitySymbol  #: Reference to the entity this architecture implements.
+	_entity:                EntitySymbol           #: Reference to the entity this architecture implements.
 
 	def __init__(
 		self,
--- a/pyVHDLModel/Namespace.py
+++ b/pyVHDLModel/Namespace.py
@@ -87,13 +87,18 @@ class Namespace(Generic[K, O]):
 	   * :class:`Concurrent declaration region <pyVHDLModel.Regions.ConcurrentDeclarationRegionMixin>`
 	   * :class:`Sequential declaration region <pyVHDLModel.Regions.SequentialDeclarationRegionMixin>`
 	"""
-	_name:            str                     #: The namespace's name.
-	_parentNamespace: "Namespace"             #: Reference to the enclosing namespace, or ``None`` for the outermost one.
-	_subNamespaces:   Dict[str, "Namespace"]  #: Dictionary of all nested namespaces, indexed by name.
-	_elements:        Dict[K, O]              #: Dictionary of all elements declared in this namespace, indexed by name.
-	_sharesRegion:    bool                    #: ``True`` if the parent namespace is the same VHDL declarative region.
+	_name:                   str                     #: The namespace's name.
+	_parentNamespace:        "Namespace"             #: Reference to the enclosing namespace, ``None`` if outermost.
+	_subNamespaces:          Dict[str, "Namespace"]  #: Dictionary of all nested namespaces, indexed by name.
+	_elements:               Dict[K, O]              #: All elements declared in this namespace, indexed by name.
+	_sharesRegionWithParent: bool                    #: ``True`` if the parent namespace is the same declarative region.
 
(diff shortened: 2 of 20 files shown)


pyVHDLModel/__init__.py
```diff
@@ -2445,6 +2446,7 @@ def LinkPackageBodies(self) -> None:
 			package._packageBody = packageBody    # TODO: add warning if package had already a body, which is now replaced
 			packageBody._package.Package = package
 			packageBody._namespace._parentNamespace = package._namespace
+			packageBody._namespace._sharesRegionWithParent = True   # one declarative region (LRM 12.1)
```
Paebbels: This should be a parameter to the initializer when the namespace gets created.

pyVHDLModel/Namespace.py
```diff
@@ -89,6 +91,7 @@ class Namespace(Generic[K, O]):
 	_parentNamespace: "Namespace"             #: Reference to the enclosing namespace, or ``None`` for the outermost one.
 	_subNamespaces:   Dict[str, "Namespace"]  #: Dictionary of all nested namespaces, indexed by name.
 	_elements:        Dict[K, O]              #: Dictionary of all elements declared in this namespace, indexed by name.
+	_sharesRegionWithParent: bool             #: ``True`` if the parent namespace is the same VHDL declarative region.
```
Paebbels: Alignment of types got broken.

pyVHDLModel/Namespace.py
```diff
@@ -101,6 +104,7 @@ def __init__(self, name: str, parentNamespace: Nullable["Namespace"] = None) ->
 		self._parentNamespace = parentNamespace
 		self._subNamespaces = {}
 		self._elements = {}
+		self._sharesRegionWithParent = False
```
Paebbels: This should be a parameter from dunder-init. For sure, it can default to False.

It also needs a read-only property.

tests/unit/Namespace/Duplicates.py
```diff
@@ -0,0 +1,209 @@
...
+	print("Use: 'python -m unitest <testcase module>'")
+	exit(1)
+
+
+def _subtype() -> SimpleSubtypeSymbol:
+	return SimpleSubtypeSymbol(SimpleName("bit"))
+
+
+def _signal(identifier: str) -> Signal:
+	return Signal([identifier], _subtype())
```
Paebbels: Could these helpers be moved to `tests/unit/__init__.py` (or another helper module)?

Please check other test modules, there should be a ton of such helpers for symbols, objects and names. The naming convention with a single underscore in front ok.

tests/unit/Namespace/Duplicates.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+def _duplicates(collector) -> list:
```
Paebbels:
```suggestion
def _duplicates(collector) -> List[DuplicateDeclarationWarning]:
```

do other tests need a similar filter? If so it could be generalized and accept a type parameter for the filtered type.

pyVHDLModel/Namespace.py
Paebbels: Applied - `Namespace.__init__` takes `sharesRegion` (defaulting to `False`) and exposes the read-only `SharesRegion`.

pyVHDLModel/Namespace.py
Paebbels: Fixed. I also shortened the field name to `_sharesRegion`, which keeps it inside the existing column.

pyVHDLModel/__init__.py
Paebbels: Applied, and it works out better than a link-time assignment. Whether a unit continues its parent's declarative region is intrinsic to the **class** - an architecture always continues its entity's - so it does not depend on when linking happens. `DesignUnit` carries it as a `ClassVar` (`_continuesParentRegion`) and passes it when creating the namespace; `Architecture` and `PackageBody` set it to `True`.

`LinkArchitectures`/`LinkPackageBodies` no longer touch the namespace flag at all, so the flag is correct from construction rather than only after linking.

tests/unit/Namespace/Duplicates.py
Paebbels: Done, and you were right that there was a ton of them. Rather than moving only the two here, I folded in **every** duplicate: 14 local definitions across 12 modules now live in `tests/unit/__init__.py` as `_subtypeSymbol`, `_entitySymbol`, `_signalSymbol`, `_variableSymbol`, `_signal`, `_variable`.

Two were the same helper under different names - `_subtype` vs `_subtypeSymbol`, and `_signalTarget`/`_variableTarget` vs the symbol helpers - so those are unified too.

The one thing that needed care: the copies had **different defaults**. `_subtype()` meant ``bit`` in `Instantiation/Interface.py` but ``natural`` in `Instantiation/Object.py`, and `_entitySymbol()` meant ``ent`` in `Instantiation/DesignUnit.py` but ``e`` elsewhere. The shared helpers take the majority default, and the call sites that relied on a different one now pass it explicitly - so no test silently changes what it builds.

Six helpers stay module-local because only one module uses them: `_architecture`, `_generic`, `_integerType`, `_port`, `_range`, `_returnType`.

tests/unit/Namespace/Duplicates.py
Paebbels: Applied and generalized: `_warningsOfType(collector, warningType)` is in `tests/unit/__init__.py`, typed `List[_Warning]` via a `TypeVar` bound to `BaseException`, so it returns the right element type for any warning class.

`Namespace/Hiding.py` had the same list comprehension inline and now uses it too. Those are the only two places filtering a collector today.

pyVHDLModel/DesignUnit.py
```diff
@@ -182,6 +182,8 @@ class DesignUnit(ModelEntity, NamedEntityMixin, DocumentedEntityMixin):
 	     * :class:`Package body <pyVHDLModel.DesignUnit.PackageBody>`
 	"""
 
+	_continuesParentRegion: ClassVar[bool] = False  #: ``True`` if it continues its parent's declarative region.
+
 	_document: 'Document'                                  #: The VHDL library, the design unit was analyzed into.
```
Paebbels, line 187: alignment of types, optional init values and comments.

pyVHDLModel/DesignUnit.py
```diff
@@ -236,7 +238,7 @@ def __init__(self, identifier: str, contextItems: Nullable[Iterable[ContextUnion
 		self._dependencyVertex = None
 		self._hierarchyVertex = None
 
-		self._namespace = Namespace(self._normalizedIdentifier)
+		self._namespace = Namespace(self._normalizedIdentifier, sharesRegion=self._continuesParentRegion)
```
Paebbels: I liked the `sharesRegionWithParent` naming. shares region doesn't express with whom.

pyVHDLModel/DesignUnit.py
```diff
@@ -650,6 +652,8 @@ class PackageBody(SecondaryUnit, DesignUnitWithContextMixin, ConcurrentDeclarati
 	   * :class:`Package it implements <pyVHDLModel.DesignUnit.Package>`
 	"""
 
+	_continuesParentRegion: ClassVar[bool] = True
+
 	_package:       PackageSymbol  #: Reference to the package this body implements.
```
Paebbels: alignment

pyVHDLModel/Namespace.py
```diff
@@ -125,6 +129,20 @@ def ParentNamespace(self, value: 'Namespace') -> None:
...
 
+	@readonly
+	def SharesRegion(self) -> bool:
+		"""
+		Read-only property to access whether this namespace continues its parent's declarative region
+		(:attr:`_sharesRegion`).
+
+		An entity and its architecture form one VHDL declarative region, as do a package and its body,
+		even though each owns a namespace. A duplicate declaration is reported across such a link, while a
+		genuinely nested region - a process, a block - hides instead.
```
Paebbels: This is a hint right? use `.. hint::`

tests/unit/Namespace/Duplicates.py
```diff
@@ -0,0 +1,201 @@
...
+region can span more than one namespace in this model - an entity and its architecture are one region,
+as are a package and its body - so the check follows those links.
+
+Overloadable declarations are exempt: several subprograms may share a name. Signatures are not compared
+yet, so any two subprograms sharing a name are accepted.
+
+The expectations below were established with the GHDL analyzer (``ghdl -a --std=08``).
+"""
+from pathlib import Path
+from unittest import TestCase
```
Paebbels, line 44: alignment of imports.

tests/unit/__init__.py
```diff
@@ -29,3 +29,53 @@
...
+
+from pyVHDLModel.Name   import SimpleName
+from pyVHDLModel.Object import Signal, Variable
+from pyVHDLModel.Symbol import EntitySymbol, SignalSymbol, SimpleSubtypeSymbol, VariableSymbol
+
+
+_Warning = TypeVar("_Warning", bound=BaseException)
+
+
+def _subtypeSymbol(name: str = "natural") -> SimpleSubtypeSymbol:
```
Paebbels: always specify the name, no default names like natural, s, e, v, ...

pyVHDLModel/DesignUnit.py
Paebbels: Agreed - reverted to `sharesRegionWithParent` for the field, the initializer parameter and the property. You are right that the short form does not say with whom.

One knock-on: the longer name widened the `Namespace` field block, which pushed `_parentNamespace` and `_elements` from 118-119 to 125-126 columns. I trimmed those two comments so the block stays within 120.

pyVHDLModel/DesignUnit.py
Paebbels: Aligned. The `#:` column now matches the field block below it in all three classes.

pyVHDLModel/DesignUnit.py
Paebbels: Aligned, and I gave this one and the `Architecture` one a `#:` comment - the `Architecture` copy had none at all.

pyVHDLModel/Namespace.py
Paebbels: Yes - changed to `.. hint::`.

tests/unit/Namespace/Duplicates.py
Paebbels: Fixed. The column was right but the order was not: `Exception` came before `Base`. Sorted, and I also aligned the `from tests.unit` import column in three other modules where it did not match the `pyVHDLModel` block.

tests/unit/__init__.py
Paebbels: Done - none of the shared helpers defaults a name any more. `_subtypeSymbol`, `_entitySymbol`, `_signalSymbol`, `_variableSymbol`, `_signal` and `_variable` all require it, and every call site across 12 modules now names what it builds.

This is better than what I had: the whole reason the previous commit needed care was that the copies disagreed on their defaults (``bit`` vs ``natural``, ``e`` vs ``ent``). With no default there is nothing to reconcile, and a test no longer hides which subtype or entity it is constructing. I added a check that no shared helper has a default and no no-arg call site remains - both report zero.

PR #161, opened 2026-07-28
Model a block statement's complete block header

Paebbels:
Findings item 4, the last of the four you ordered.

⚠️ **Needs the companion pyGHDL PR**: Paebbels/GHDL#17.

## The gap

A block statement could only model its **port clause**. Its generic clause and both map aspects were dropped, so this legal VHDL lost three quarters of its header:

```vhdl
blk : block
    generic (G : positive := 1);
    generic map (G => 2);
    port (bp : in bit);
    port map (bp => clock);
begin
end block;
```

## The TODOs were the fix

`GenericMapAspectMixin` and `PortMapAspectMixin` are new in `Association.py`. That resolves the `# TODO: extract to mixin` markers you had left - the association fields were hand-rolled in **four** places, not just the two in `Instantiation`:

| Class | Was |
|---|---|
| `Instantiation` | both fields, both marked TODO |
| `PackageInstantiation` | generic only, marked TODO |
| `SubprogramInstantiationMixin` | generic only |
| `BindingIndication` | both fields |

All four now use the mixins, and `ConcurrentBlockStatement` gains both plus `WithGenericsMixin`.

The mixins type their fields as `GenericAssociationItem`/`PortAssociationItem` rather than the base `AssociationItem` that `Instantiation` used. I checked pyGHDL first: it already constructs the specific types, so nothing downstream loosens.

## A wrong test this surfaced

`test_WithPortItems` passed a **`PortAssociationItem` as a port item** - an association where an interface item belongs. It only passed because nothing type-checks that argument. Fixed to a `PortSimpleSignalInterfaceItem`, and joined by coverage for the generic clause, both map aspects, and a block with no header at all.

## Verification

| Check | Result |
|---|---|
| `pytest tests/unit` | **446** passed (+4), 69 subtests |
| pyGHDL `testsuite/pyunit` (with the companion PR) | 185 passed, 17 xfailed |
| Round-trip on real VHDL: block with full header | all four lists populated |
| Block with partial / no header | correct, no crash |
| `# TODO: extract to mixin` remaining | **0** (was 3) |
| ReST / caret alignment | 0 problems |
| New lines > 120 chars | 0 |

### DIFF
--- a/pyVHDLModel/Configuration.py
+++ b/pyVHDLModel/Configuration.py
@@ -204,13 +204,12 @@ class BindingIndication(ModelEntity, GenericMapAspectMixin, PortMapAspectMixin):
 		:param parent:                  The parent model entity of this entity.
 		"""
 		super().__init__(parent)
+		GenericMapAspectMixin.__init__(self, genericAssociationItems)
+		PortMapAspectMixin.__init__(self, portAssociationItems)
 
 		self._entityAspect = entityAspect
 		if entityAspect is not None:
 			entityAspect.Parent = self
-
-		GenericMapAspectMixin.__init__(self, genericAssociationItems)
-		PortMapAspectMixin.__init__(self, portAssociationItems)
 
 	@readonly
 	def EntityAspect(self) -> Nullable[EntityAspect]:
--- a/pyVHDLModel/Configuration.py
+++ b/pyVHDLModel/Configuration.py
@@ -44,6 +44,7 @@ from pyVHDLModel.Name        import Name
 from pyVHDLModel.Symbol      import Symbol, EntitySymbol, ArchitectureSymbol, ConfigurationSymbol
 from pyVHDLModel.Symbol      import ComponentInstantiationSymbol
 from pyVHDLModel.Association import GenericAssociationItem, PortAssociationItem
+from pyVHDLModel.Association import GenericMapAspectMixin, PortMapAspectMixin
 
 
 @export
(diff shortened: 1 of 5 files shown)


pyVHDLModel/Configuration.py
```diff
@@ -210,17 +209,8 @@ def __init__(
...
-				self._genericAssociationItems.append(association)
-				association.Parent = self
-
-		self._portAssociationItems = []
-		if portAssociationItems is not None:
-			for association in portAssociationItems:
-				self._portAssociationItems.append(association)
-				association.Parent = self
+		GenericMapAspectMixin.__init__(self, genericAssociationItems)
+		PortMapAspectMixin.__init__(self, portAssociationItems)
```
Paebbels: Why not called after super()....?

pyVHDLModel/Instantiation.py
```diff
@@ -271,16 +258,11 @@ def __init__(
 		super().__init__(identifier, contextItems, documentation=documentation, parent=parent)
 		GenericEntityInstantiationMixin.__init__(self)
 
+		GenericMapAspectMixin.__init__(self, genericAssociationItems)
```
Paebbels, line 259:
```suggestion
		GenericEntityInstantiationMixin.__init__(self)
		GenericMapAspectMixin.__init__(self, genericAssociationItems)
```

pyVHDLModel/Configuration.py
Paebbels: No good reason - fixed. `BindingIndication` now calls both mixin initializers directly after `super().__init__(parent)`, before setting its own `_entityAspect`, which is the shape `Instantiation` already used.

pyVHDLModel/Instantiation.py
Paebbels: Applied. `SubprogramInstantiationMixin` had the same stray blank line between `super()` and the mixin call, so I grouped that one too - all four converted sites now read alike: `super()`, then every mixin initializer, then the class's own fields.

I added a check over the four classes asserting the mixin initializer appears within three statements of `super()`; all pass.

PR #162, opened 2026-07-29
Show the interface mode when rendering an interface item

Paebbels:
Findings item: the follow-up left open by #157.

## The gap

An interface item inherits `Obj.__str__`, because `PortSimpleSignalInterfaceItem` and friends descend from `Signal`/`Constant`/`Variable`/`File`. That renders `p : bit` where VHDL writes `p : in bit` - the **mode is lost**, and an object's rendering cannot carry it, since a mode is not part of an object declaration.

## Now

```
signal p1, p2 : in bit          constant G : in positive
variable v : inout integer      signal s : out bit
file f : text                   signal p : view myView
```

* `Mode.Default` means the mode is context dependent, so nothing is rendered for it.
* The two VHDL-2019 **mode-view** items keep their mode view in the *subtype position* (`PortViewSignalInterfaceItem.__init__` passes it as the subtype), so they are marked `view` rather than pretending it is a subtype.
* `ParameterFileInterfaceItem` declares no mode and simply renders as the object.

## On the convention

`formatInterfaceItem()` keeps the five moded variants from repeating the same three lines. It is a **module-level helper** - following `identifiersOf` in `Base.py` - and not a mixin method, since representation belongs to the primary inheritance tree. The check added in #157 still reports **0 mixins defining `__str__`/`__repr__`**.

## Verification

| Check | Result |
|---|---|
| `pytest tests/unit` | **452** passed (+6), 69 subtests |
| Mixins defining `__str__`/`__repr__` | 0 |
| Interface group rendering | unchanged (groups list names via `identifiersOf`) |
| ReST parse | 0 problems |
| New lines > 120 chars | 0 |

The new tests expect a trailing `?` on each subtype - that is `Symbol.__str__`'s unresolved marker, since the items are built standalone and their subtype symbols are never linked. Noted in the test class doc-string so it is not mistaken for a typo.

### DIFF
--- a/pyVHDLModel/Interface.py
+++ b/pyVHDLModel/Interface.py
@@ -244,6 +244,26 @@ class ModeViewDeclaration(ModelEntity, NamedEntityMixin, DocumentedEntityMixin):
 
 
 @export
+def formatInterfaceItem(item, mode: Nullable[Mode] = None, isModeView: bool = False) -> str:
+	"""
+	Format an interface item as a single-line declaration.
+
+	Interface items are objects, so they render like one (``signal p : bit``), but an object's rendering
+	cannot show the interface *mode* - it is not part of an object declaration. This adds it.
+
+	:param item:        The interface item to format.
+	:param mode:        The item's mode, or ``None`` when it declares none.
+	:param isModeView:  ``True`` if the subtype position holds a mode view (VHDL-2019) rather than a subtype.
+	:returns:           Formatted interface item.
+	"""
+	prefix = "" if mode is None or mode is Mode.Default else f"{mode!s} "
+	if isModeView:
+		prefix = "view "
+
+	return f"{item._objectKeyword} {', '.join(item._identifiers)} : {prefix}{item._subtype}"
+
+
+@export
 class InterfaceItemMixin(metaclass=ExtendedType, mixin=True):
 	"""
 	A mixin-class marking a declaration as an interface item.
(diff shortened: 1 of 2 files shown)


pyVHDLModel/Interface.py
```diff
@@ -243,6 +243,26 @@ def __str__(self) -> str:
 		return f"view {self._identifier} of {self._subtype}: {elements}"
 
 
+@export
+def formatInterfaceItem(item, mode: Nullable[Mode] = None, isModeView: bool = False) -> str:
```
Paebbels: Why is it not a method in one of the many mixins we have? E.g. we have `InterfaceItemMixin` can't we share it like that? If yes, don't write standalone functions if we have a class hierarchy to share code and operations with matching types.

Paebbels: Applied - it is `InterfaceItemMixin._FormatInterfaceItem()` now, so the shared logic travels with the hierarchy rather than sitting beside it.

For the record on why I reached for a free function: I read the #156 rule ("a mixin is not responsible for the external view or representation of the class") as covering the formatting helper too. Your framing there was more precise than my reading - a mixin brings *fields, properties and maybe methods (algorithms)*, and this is an algorithm. What must stay off the mixin is `__str__` itself, and it has: each concrete interface item still defines its own, and the check from #157 still reports 0 mixins defining `__str__`/`__repr__`.

PR #163, opened 2026-07-29
A parenthesis name may also be a type conversion

Paebbels:
Companion to Paebbels/GHDL#18, from investigating why `TypeConversion` is never constructed.

`IndexedObjectOrFunctionCallSymbol` declared `Object | Function`. But the syntax it represents has a **third** reading: `arr(0)`, `f(0)` and `integer(0)` are written identically, so a parser cannot tell an indexed name from a function call from a **type conversion**.

`Type` and `Subtype` join its possible references, and the doc-string names all three readings.

## Why this matters beyond tidiness

GHDL's parser emits `Parenthesis_Name` for all three; only semantic analysis rewrites one into `Type_Conversion`. `pyGHDL.dom` parses without that pass, so **this symbol is what a type conversion actually becomes today** - and it was under-declaring that. Anything filtering on `_possibleReferences` would have excluded a legitimate resolution target.

| Check | Result |
|---|---|
| `pytest tests/unit` | 446 passed, 69 subtests |
| ReST parse | 0 problems |
| New lines > 120 chars | 0 |

The existing test asserted the old pair and is updated, with a note on why all three are carried.

### DIFF
--- a/pyVHDLModel/Symbol.py
+++ b/pyVHDLModel/Symbol.py
@@ -1071,15 +1071,24 @@ class SimpleObjectOrFunctionCallSymbol(Symbol):
 @export
 class IndexedObjectOrFunctionCallSymbol(Symbol):
 	"""
-	Represents a reference that is either an indexed object or a function call.
+	Represents a reference that is an indexed object, a function call or a type conversion.
 
-	An expression like ``f(0)`` may index an array or call a function; both look identical until the
-	name is resolved. The referenced language entity is available as :data:`Reference` once resolved.
+	All three are written the same way - ``arr(0)``, ``f(0)`` and ``integer(0)`` are indistinguishable
+	as syntax, so a parser produces one shape for them and only name resolution tells them apart. The
+	referenced language entity is available as :data:`Reference` once resolved.
+
+	.. seealso::
+
+	   * :class:`Type conversion <pyVHDLModel.Expression.TypeConversion>`
+	   * :class:`Simple object or function call <pyVHDLModel.Symbol.SimpleObjectOrFunctionCallSymbol>`
 	"""
 	def __init__(self, name: Name) -> None:
 		"""
-		Initializes a reference that is either an indexed object or a function call.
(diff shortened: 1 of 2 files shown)


pyVHDLModel/Symbol.py
```diff
@@ -1071,15 +1071,24 @@ def __init__(self, name: Name) -> None:
 @export
 class IndexedObjectOrFunctionCallSymbol(Symbol):
 	"""
-	Represents a reference that is either an indexed object or a function call.
+	Represents a reference that is an indexed object, a function call or a type conversion.
```
Paebbels: The phrasing `is an ...` isn't as strong as `is either an ...`. Is my wording not good?

pyVHDLModel/Symbol.py
```diff
@@ -1071,15 +1071,24 @@ def __init__(self, name: Name) -> None:
...
 
-	An expression like ``f(0)`` may index an array or call a function; both look identical until the
-	name is resolved. The referenced language entity is available as :data:`Reference` once resolved.
+	All three are written the same way - ``arr(0)``, ``f(0)`` and ``integer(0)`` are indistinguishable
+	as syntax, so a parser produces one shape for them and only name resolution tells them apart. The
+	referenced language entity is available as :data:`Reference` once resolved.
```
Paebbels: Should we mark this as an `.. attention::` ?

pyVHDLModel/Symbol.py
Paebbels: Your wording is better and I have restored it, in both the class and the initializer doc-string. "is either an ..." states the exclusivity - exactly one of the three - whereas "is an ..." leaves it open. I had changed it only because "either" is conventionally a two-way word, but that is a weak reason to lose the meaning.

pyVHDLModel/Symbol.py
Paebbels: Yes - changed to `.. attention::`. It reads better separated from the summary: the first line says what the class represents, and the admonition explains why the three readings cannot be told apart.

PR #164, opened 2026-07-29
Let a record element carry documentation

Paebbels:
Fixes the gap found while reviewing Paebbels/GHDL#19.

`RecordTypeElement(ModelEntity, MultipleNamedEntityMixin)` had **no `DocumentedEntityMixin`**, so its initializer was `(identifiers, subtype, parent)` with no `documentation` slot. A doc comment on a record element had nowhere to go:

```vhdl
type frame is record
    --! The first field.
    a : bit;
end record;
```

Every other declaration-like class in the model accepts one - `Attribute`, `Alias`, `Obj`, `BaseType`, `Subprogram`, the design units - so this was an oversight rather than a decision. The mixin and a `documentation` parameter are added, and the class doc-string's example gains the comment line it can now model.

## Positional compatibility

`documentation` goes **before** `parent`, matching every other class in the model. I checked every construction site first: only this repo's tests and pyGHDL build one, and neither passed `parent` positionally, so nothing breaks.

## Companion

Paebbels/GHDL#19 extracts the comment. End-to-end on real source:

```
record   'frame': '--! A documented record.'
element  ('a',): '--! The first field.'
element  ('b', 'c'): '--! The second and third fields.'
```

## Other classes with the same gap

A survey found nine named classes without `DocumentedEntityMixin`. Four are libraries (`Library`, `Ieee`, `Std`, `PredefinedLibrary`) - not source declarations, correctly excluded. The remaining four are genuine candidates I have **not** touched, since each changes another public constructor:

* `ModeViewElement`, `SimpleModeViewElement`, `CompositeModeViewElement` - element declarations inside a mode view, exactly analogous to a record element
* `DefaultClock` - a PSL declaration

Say the word and I will do those too.

| Check | Result |
|---|---|
| `pytest tests/unit` | **453** passed (+1), 69 subtests |
| pyGHDL `testsuite/pyunit` (with the companion) | 187 passed, 17 xfailed, 12 subtests |
| ReST / caret alignment / example round-trip | 0 problems |
| New lines > 120 chars | 0 |

### DIFF
--- a/pyVHDLModel/Interface.py
+++ b/pyVHDLModel/Interface.py
@@ -49,7 +49,7 @@ from pyVHDLModel.Type       import Type
 
 
 @export
-class ModeViewElement(ModelEntity, MultipleNamedEntityMixin):
+class ModeViewElement(ModelEntity, MultipleNamedEntityMixin, DocumentedEntityMixin):
 	"""
 	Base-class for one element definition inside a mode view declaration (VHDL-2019). An element may name
 	several fields sharing the same specification (e.g. ``a, b : out;``), hence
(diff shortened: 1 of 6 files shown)


Paebbels:
Extended to the remaining three, as agreed.

`ModeViewElement` gains the mixin, so `SimpleModeViewElement` and `CompositeModeViewElement` both inherit it (each threads `documentation` through its own initializer), and `DefaultClock` gains it too.

**All four remaining classes without `DocumentedEntityMixin` are now libraries** - `Library`, `Ieee`, `Std`, `PredefinedLibrary` - which correctly have none: a library is not a source declaration and has no comment to carry. A check confirms nothing else is left.

`DefaultClock` had **no test at all**, so it has two now.

The pyGHDL companion (Paebbels/GHDL#19) extracts all of it:

```
view InnerView of InnerRecord is
    --! An output element.
    x : out;
    --! An input element.
    y : in;
end view;
```

One thing that surfaced: `Mode_View_Declaration` is affected by the **same GHDL comment-attachment quirk** already recorded for `Anonymous_Type_Declaration` and `Object_Alias_Declaration` - the declaration itself gets no comment, and its first element receives both. The extraction is correct; the attachment is upstream. I have added it to that findings entry.

| Check | Result |
|---|---|
| `pytest tests/unit` | **458** passed (+6), 69 subtests |
| pyGHDL `testsuite/pyunit` | 188 passed, 17 xfailed, 12 subtests |
| Named classes still without `DocumentedEntityMixin` | 4, all libraries |
| ReST / caret alignment / round-trip | 0 problems |
| `interrogate` | 97.4% |
| New lines > 120 chars | 0 |

PR #165, opened 2026-07-29
Let a library carry documentation, settable after construction

Paebbels:
VHDL has no library declaration, so there is nowhere in the source to hang a comment - but a caller of this API may well have one from elsewhere: a compile-order file, a project description, a build system.

## What

`Library` gains `DocumentedEntityMixin` and a `documentation` parameter, and **overrides `Documentation` as a settable property**:

```py
library = Library("mylib", "--! From a compile-order file.")
library.Documentation = "--! Supplied later."
```

The setter is the point of the change, and it is deliberately different from every other documented entity: those take their text from the source and keep it read-only, whereas a library's can only ever come from outside - and may not be known when the library is constructed. A `.. hint::` on the property records that reasoning.

`PredefinedLibrary` accepts and forwards one too. `Std` and `Ieee` pass **none**, since VHDL defines no documentation for the predefined libraries, but a caller can still set it afterwards:

```
Ieee().Documentation        -> None
std.Documentation = "..."   -> works
```

## Parameter position

`documentation` sits **before** `allowBlackbox`, matching `Package` and the rest of the model rather than being appended for convenience.

That does change what a positional second argument means. I checked every construction site: the only positional callers were **tests passing `None`**, so behaviour is identical either way - but leaving them would have meant a positional that silently changed meaning, so they now pass nothing instead. Nothing outside this repo constructs a `Library` positionally; pyGHDL never constructs one at all.

## Verification

| Check | Result |
|---|---|
| `pytest tests/unit` | **455** passed (+3), 69 subtests |
| pyGHDL `testsuite/pyunit` (its base branch, to isolate this change) | 185 passed, 17 xfailed |
| Ambiguous positional `Library(..., None)` calls left | 0 |
| ReST parse | 0 problems |
| New lines > 120 chars | 0 |

Independent of #164 - different classes, no conflict.

### DIFF
--- a/pyVHDLModel/__init__.py
+++ b/pyVHDLModel/__init__.py
@@ -2198,7 +2198,7 @@ class Design(ModelEntity, AllowBlackboxMixin):
 
 
 @export
-class Library(ModelEntity, NamedEntityMixin, AllowBlackboxMixin):
+class Library(ModelEntity, NamedEntityMixin, DocumentedEntityMixin, AllowBlackboxMixin):
 	"""
 	A ``Library`` represents a VHDL library. It contains all *primary* and *secondary* design units.
 
@@ -2220,6 +2220,7 @@ class Library(ModelEntity, NamedEntityMixin, AllowBlackboxMixin):
 	def __init__(
 		self,
 		identifier:    str,
+		documentation: Nullable[str] = None,
 		allowBlackbox: Nullable[bool] = None,
 		parent:        Nullable[ModelEntity] = None
 	) -> None:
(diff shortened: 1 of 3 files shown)


Paebbels:
Rebased onto `dev` after #164 - conflict resolved and force-pushed (`cb35422`). GitHub now reports **mergeable**.

The conflict was small: both branches appended an import to `tests/unit/Instantiation/Model.py` - #164 added `DefaultClock`, this one added `Std`. Both are kept, and I realigned the whole `from pyVHDLModel.` block so the `import` column is uniform again rather than leaving the two new lines a character out.

Re-verified after the rebase, since the two changes touch the same area:

| Check | Result |
|---|---|
| `pytest tests/unit` | **461** passed, 69 subtests |
| pyGHDL `testsuite/pyunit` (merged `paebbels/dom-updates`, incl. #19) | 188 passed, 17 xfailed, 12 subtests |
| Named classes that **cannot** be documented | **0** |
| ReST parse | 0 problems |
| `interrogate` | 97.4% |
| New lines > 120 chars | 0 |

With #164 and #19 merged, that last row is the milestone: every named class in the model can carry documentation, and pyGHDL extracts it wherever VHDL can express it.

pyVHDLModel/__init__.py
```diff
@@ -2243,6 +2246,25 @@ def __init__(
...
 
+	@property
+	def Documentation(self) -> Nullable[str]:
+		"""
+		Property to access the library's documentation (:attr:`_documentation`).
+
+		.. hint::
+
+		   Unlike every other documented entity, a library's documentation cannot come from VHDL source:
+		   the language has no library declaration to hang a comment on. It is therefore settable, so a
```
Paebbels:
```suggestion
		   the language has no library declaration to attach a comment to. It is therefore settable, so a
```

PR #166, opened 2026-07-31
Declare `_FORMAT` and `_direction` as `ClassVar`, require pyTooling 8.18

Paebbels:
Fixes the release failure: every job of the v0.38.0 run on `main` failed, and `dev` fails the same way now that **pyTooling v8.18.0 is on PyPI**. CI installs pyTooling from PyPI, so the checkout was fine until the release landed.

```
pyTooling.MetaClasses.DuplicateFieldInSlotsError:
  Slot '_FORMAT' is shadowed by a class member in class 'NegationExpression'.
```

28 collection errors, so **0 tests ran** — on every OS and every Python version.

## The defect

`UnaryExpression`, `BinaryExpression`, `TernaryExpression` and `RangeExpression` declare `_FORMAT` / `_direction` as *annotated instance fields*, so `ExtendedType` turns them into slots. 48 derived classes then assign them as constants in the class body:

```py
class UnaryExpression(BaseExpression):
	_FORMAT:  Tuple[str, str]  #: ...            <- becomes a slot

class NegationExpression(UnaryExpression):
	_FORMAT = ("-", "")                          <- class attribute hiding the slot descriptor
```

This is a **long-standing defect that v8.18.0 surfaced**, not a new rule. The assignment creates a class attribute that hides the slot's descriptor: reading the field works, assigning it on an instance raises `AttributeError`. `TernaryExpression._FORMAT` already carried

```py
# FIXME: needs ClassVar[...] when pyTooling gets fixed.
```

pyTooling is now fixed — v8.18.0 stopped treating a `ClassVar` without an initial value as a slot — so this is that FIXME being resolved.

## The fix

Nothing assigns either field per instance; both are constants of the derived class. So they are annotated `ClassVar[...]` at the four declarations and at all 48 assignments:

```py
	_FORMAT:  ClassVar[Tuple[str, str]]  #: ...
	...
	_FORMAT: ClassVar[Tuple[str, str]] = ("-", "")
```

The parameterisation is derived from each tuple's arity — `Tuple[str, str]` for unary, `Tuple[str, str, str]` for binary and range, `Tuple[str, str, str, str]` for ternary — rather than widened to a single loose type.

## The dependency floor has to move

`pyTooling ~= 8.17` → `~= 8.18`, and this is **not cosmetic**. Under 8.17.0 a `ClassVar` without an initial value still became a slot, so the corrected declaration fails there too, with the older message:

```
AttributeError: Slot '_FORMAT' declared in class 'NegationExpression'
                already exists in base-class 'pyVHDLModel.Expression.UnaryExpression'.
```

There is no formulation that satisfies both releases, so 8.18 is a genuine floor. Updated in `requirements.txt`, `pyproject.toml` and all three `doc/Dependency.rst` tables — the grid tables keep their column widths (checked: 6 tables, 0 with mixed row widths).

## Verification

Against pyTooling 8.18+ (`PYTHONPATH=/work/pyTooling`, i.e. v8.18.0 plus the merged property work):

| Check | Before | After |
|---|---|---|
| `pytest tests/unit` | **28 collection errors, 0 tests run** | **461 passed** + 69 subtests |
| Fresh import of every module, `UnannotatedFieldWarning` | — | **0** |
| `_FORMAT` / `_direction` in any `__slots__` | yes | **no** |
| Field-comment alignment (`Expression.py`) | — | 0 misaligned |
| Annotations forced to evaluate (PEP 649, simulates 3.11–3.13) | — | clean |

`__str__` renders unchanged for all four shapes:

```
 negation      : -3
 addition      : 1 + 2
 ascending rng : 0 to 3        | direction: to
 descending rng: 3 downto 0    | direction: downto
 when-else     : 1 when 0 else 2
```


Paebbels:
## CI result: 52 passed, 10 skipped, 1 failed

The unit tests are fixed — **52 jobs green** where the v0.38.0 release run had 0 tests running on every OS and Python version.

The one remaining failure is **not caused by this PR**, and it needs action in two other repositories. Both were previously *masked*: on `dev` the Documentation job dies earlier, at `Artifact not found for name: pyVHDLModel-UnitTestReportSummary-XML`, because no unit test ever produced a report. Fixing the tests is what let the job get far enough to fail properly.

### Blocker 1 — `sphinx_reports` pins a broken `pyEDAA.Reports`

```
File ".../sphinx_reports/DataModel/CodeCoverage.py", line 51
  from pyEDAA.Reports.DocumentationCoverage.Python import PackageCoverage
File ".../pyEDAA/Reports/DocumentationCoverage/Python.py"
  class DocStrCoverage(metaclass=ExtendedType):
  	del ResultCollection
...
  _coverageReport:  ResultCollection
NameError: name 'ResultCollection' is not defined
```

`sphinx_reports 0.11.1` requires `pyEDAA.Reports~=0.17.4`, i.e. `>=0.17.4,<0.18`. The `del ResultCollection` defect was fixed in **pyEDAA.Reports v0.18.0**, which is on PyPI since 06:13 today — but that pin makes it unreachable. `sphinx-reports` `dev` still carries `pyEDAA.Reports ~= 0.17.4`.

Worth being precise about the trigger, because it changes who has to act: this is **pyTooling v8.18.0 exposing it**, not a coincidence. HTML Documentation last succeeded on 2026-07-29; pyTooling 8.18.0 was published 2026-07-30 21:37. Under 8.17.0 the non-slots branch of `_computeSlots` had no PEP 649 fallback and read only `members["__annotations__"]` — absent on Python 3.14 — so it never evaluated the annotation and the defect stayed asleep. 8.18.0 fixed that branch, so it now evaluates `_coverageReport: ResultCollection` and raises.

**Fix:** release `sphinx_reports` with `pyEDAA.Reports ~= 0.18`. Nothing here can work around it — pinning `pyEDAA.Reports >= 0.18` in `doc/requirements.txt` just makes pip fail on the conflict.

### Blocker 2 — `VersionCheck` rejects the `v` prefix

From the v0.38.0 release run:

```
expected:  v0.38.0
from code: 0.38.0
##[error]Expected version (v0.38.0) doesn't match the version in Python code (0.38.0).
```

In `pyTooling/Actions` `CompletePipeline.yml@r7`:

```py
expectedVersion = "${{ needs.Prepare.outputs.version }}".strip()   # -> "v0.38.0", from the tag
versionInformation = extractVersionInformation(versionFile)        # .Version -> "0.38.0", a plain str
if expectedVersion != versionInformation.Version:                  # str vs str -> mismatch
```

`extractVersionInformation(...).Version` is a `str` and never carries a prefix, while `Prepare.outputs.version` comes from the tag and this repo tags `v0.37.0`, `v0.38.0`. So the two can never be equal, and the job fails for **every** release of a `v`-tagged repo using `CompletePipeline.yml`. Checked against both pyTooling 8.17.0 and 8.18.0 — identical behaviour, so it is not a pyTooling regression.

A one-line fix is available, since pyTooling's own version type already equates the two:

```py
>>> "v0.38.0" != SemanticVersion.Parse("0.38.0")
False
```

so comparing `SemanticVersion.Parse(expectedVersion) != SemanticVersion.Parse(versionInformation.Version)` — or simply stripping a leading `v`/`r` — resolves it.

I can open PRs for either or both if you want them.

Paebbels:
Added the pickling regression test the findings entry called for — `_FORMAT`/`_direction` being slots is what made **any model containing an expression unpicklable**, since `pickle`'s default `__setstate__` writes every slot back and hit the read-only descriptor. That is the consequence worth locking down, not just that construction succeeds.

New `ClassVariables` test class (`tests/unit/Instantiation/Expression.py`):

* `test_NotASlot` — walks the MRO of one class per arity, asserts neither name appears in any `__slots__`.
* `test_Picklable` — round-trips unary, binary, both ranges and ternary through `pickle`.
* `test_DirectionIsPerClass` — the two range expressions carry their direction as a class constant, readable without an instance.

**461 → 464 passed, 69 → 82 subtests.**

One note on the shape of the fix: the findings entry expected a judgement call about what neutral default `RangeExpression._direction` should carry on an abstract base. There is none to make — the base declares `_direction: ClassVar[Direction]` with **no value**, as a forward declaration for the subclasses to fill in. That is precisely the pyTooling behaviour v8.18.0 introduced, and it is also why the `~= 8.18` floor is hard.

PR #167, opened 2026-07-31
v0.38.0

Paebbels:
> [!NOTE]
> This release accumulated 32 pull-requests over three months, so the notes below are grouped by theme
> rather than enumerating individual classes. The linked pull-requests carry the per-class detail.

The model grew from **384 to 433 classes**; the package is **+15,743 / -985 lines**.

# New Features

* **VHDL-2019 mode views.** `ModeViewDeclaration`, `ModeViewElement` and the symbols addressing them, plus the split of the port and parameter signal interface items into a *simple* and a *mode view* variant, so a port declared `view v` is no longer forced through the subtype position. (#134)
* **VHDL-2008 subprogram instantiation and aliases.** Generic subprograms can be instantiated, and `Alias` models alias declarations — with and without a subtype indication, which is what decides whether the alias may name anything or only an object. (#135)
* **Configuration and binding model.** A new `pyVHDLModel.Configuration` module: configuration declarations, block/component configurations, binding indications and the entity/configuration aspects. This was the last design unit with no model at all. (#137)
* **Complete assignment statements.** Conditional and selected assignments (concurrent and sequential), plus signal force and release. (#138)
* **Ranges are no longer one class.** `Range` is split into `SimpleRange` (`a to b`) and `RangeFromName` (a range named by an object or subtype), with `RangeAttributeSymbol` for `'range`/`'reverse_range`. The single class could not represent a range that isn't written as two bounds. (#145)
* **Block headers.** A block statement carries its complete header — generic clause, generic map, port clause and port map — rather than only its declarative and statement parts. (#161)
* **Declarative regions and namespaces.** Sequential statements, protected types and protected type bodies are declaration regions; generics, ports and parameters live in namespaces of their own; and a declarative region now **detects duplicate declarations** and reports them as a warning. (#148, #159, #160)
* **Type conversions.** A parenthesis name may be a type conversion, not only an indexed or sliced name. (#163)
* **Rendering.** The primary base classes and the interface items implement `__str__`, and an interface item shows its **mode** (`signal p : in bit`) — an object's own rendering cannot, since a mode is not part of an object declaration. (#157, #162)
* **Documentation on more entities.** Record elements carry documentation, and a library carries documentation that is **settable after construction** — VHDL has no library declaration to hang a comment on, so it can only come from outside the source. (#164, #165)

# Changes

* **The minimum pyTooling version is now 8.18** (`~= 8.17` → `~= 8.18`), in `requirements.txt`, `pyproject.toml` and `doc/Dependency.rst`. This is a **hard floor, not a tidy-up**: the `ClassVar` declarations required by the bug fix below do not work under 8.17.0 either, because a `ClassVar` without an initial value still became a slot there. There is no formulation compatible with both releases. (#166)
* **Association lists renamed for consistency:** `genericAssociations`, `portAssociations` and `parameterMappings` are now `genericAssociationItems`, `portAssociationItems` and `parameterAssociationItems`. This is a **breaking rename** of constructor parameters and properties. (#128)
* **Exception pattern.** The module's exceptions follow one pattern, low-level exceptions are chained rather than swallowed, and messages carry a solution as an exception note. (#149)
* **`@readonly` throughout.** Every property without a setter is marked `@readonly` rather than `@property`. (#149, #150)
* **Boolean naming unified** (`Is…`/`Has…`), and the duplicate class names in `IEEE.py` are de-duplicated. (#155)

# Bug Fixes

Roughly a dozen fixes, most of them crashes on real VHDL that only surfaced once pyGHDL.dom fed the model actual designs:

* **Constructor gaps and crashes:** `Function.ReturnType`, the `Subprogram` generic/parameter item gap (#129), and `GenericProcedureInterfaceItem`/`GenericFunctionInterfaceItem` on real VHDL-2008 generic subprogram clauses — fixed in #143 and again in #144 after a regression.
* **`ConstrainedScalarSubtypeSymbol` had nowhere to put its range constraint**, so the constraint was silently dropped. (#136)
* **Four expression/interface/region gaps** found by the level-1 test sweep. (#141)
* **`NameError` on Python < 3.14** from an annotation left stale by the `Range` split. Python 3.14 defers annotation evaluation ([PEP 649](https://peps.python.org/pep-0649/)), so this was invisible locally and failed on three of the four CI Python versions. (#146)
* **Two `__str__` defects** producing malformed VHDL. (#155)
* **`_FORMAT` and `_direction` were slots, not class variables.** `UnaryExpression`, `BinaryExpression`, `TernaryExpression` and `RangeExpression` declared them as annotated instance fields, so `ExtendedType` made them slots, and the 48 concrete subclasses then assigned them in the class body — which hides the slot descriptor. The field could be read but never assigned on an instance, and since `pickle`'s default `__setstate__` writes every slot back, **any model containing an expression was unpicklable**. Both are now `ClassVar[...]`. (#166)

# Documentation

This is the largest part of the release. `pyVHDLModel` went from partially documented to fully documented:

* **Every class has a doc-string**, and each one carries a **VHDL example** showing the construct it models, with `^^^` markers naming the parts that map onto the class's fields. (#151)
* **Every class field carries a `#:` comment** — 204 added, taking fields from 78/282 (27.7 %) to **282/282 (100 %)**. interrogate does not count fields at all, so this closed a gap that was invisible to the coverage number. (#152)
* **All 246 `__init__` methods** are documented, with the `:param:` descriptions derived from those field comments so the two stay consistent by construction. Doing the fields first is what made this pass mechanical. (#153)
* **All 75 dunder methods** are documented. (#154)
* **`.. seealso::` cross-references** link each base class to its derived classes.
* **interrogate: 33.7 % → 97.4 %**, and `fail-under` is raised **59 → 90** so the gain cannot silently regress. Nested helper functions are excluded (`ignore-nested-functions`); they are implementation-internal and never rendered by Sphinx.
* **`docstr_coverage`: 33.71 % → 96.77 %**, undocumented items 706 → 40.

Three pre-existing documentation bugs were found by checking the generated text against the code rather than by reading it: `InterfaceGroup.__init__` was titled for the wrong class and documented none of its parameters, `BaseType.__init__` was missing a parameter, and `Design.__init__` listed its parameters in the wrong order.

# Unit Tests

* **60 → 464 tests** (plus 82 subtests).
* **Level-1 construction/property/parent-wiring tests across the whole model** — every class is instantiated, every property read back, and every child's `Parent` verified. This sweep is what found most of the bugs listed above. (#139)
* **Namespace tests** and a dedicated `Instantiation` test package. (#147)
* Duplicate-declaration detection has its own test package. (#160)
* A `ClassVariables` test class asserts what the shadowed slots actually broke: neither name appears in any `__slots__` along the hierarchy, an expression of every arity round-trips through `pickle`, and the two range expressions carry their direction as a class constant. (#166)

----------
# Related Issues and Pull-Requests

* #128 
* #129 
* #134 
* #135 
* #136 
* #137 
* #138 
* #139 
* #140 
* #141 
* #142 
* #143 
* #144 
* #145 
* #146 
* #147 
* #148 
* #149 
* #150 
* #151 
* #152 
* #153 
* #154 
* #155 
* #157 
* #159 
* #160 
* #161 
* #162 
* #163 
* #164 
* #165
* #166

Companion work in **pyGHDL.dom** (`Paebbels/GHDL`, branch `paebbels/dom-updates`) translates GHDL's IIR tree into the classes added here; it depends on this release being published to PyPI.


Paebbels:
**Do not merge yet — this will fail exactly as the last attempt did.**

`dev` still carries the `_FORMAT`/`_direction` shadowed-slot defect, so with pyTooling v8.18.0 now on PyPI every job fails at collection (28 errors, 0 tests run). #166 fixes it and has to land on `dev` first.

The description above is the one from #133, unchanged as requested. Two things it does not yet cover, both coming from #166 — say the word and I will fold them in before you merge:

* **Dependency floor moves to `pyTooling ~= 8.18`** (from `~= 8.17`), in `requirements.txt`, `pyproject.toml` and `doc/Dependency.rst`. It is a genuine floor, not a bump for tidiness: the corrected `ClassVar` declaration does not work under 8.17.0 either.
* **A bug fix**: `_FORMAT` and `_direction` were annotated instance fields that 48 derived classes shadowed with class constants, so the field could be read but never assigned on an instance. `TernaryExpression._FORMAT` carried a `# FIXME: needs ClassVar[...] when pyTooling gets fixed.` for exactly this.

Merge order: **#166 → `dev` → this PR**.

PR #172, opened 2026-08-04
Require a range constraint in ConstrainedScalarSubtypeSymbol

Paebbels:
## Changes

`ConstrainedScalarSubtypeSymbol`'s range constraint is now mandatory.

VHDL's grammar leaves no state in which this class has no constraint:

```
subtype_indication ::= [ resolution_indication ] type_mark [ constraint ]
constraint         ::= range_constraint | array_constraint | record_constraint
range_constraint   ::= **range** range
range              ::= range_attribute_name | simple_expression direction simple_expression
```

The `range` keyword is always followed by a range, so a subtype indication is either a bare type mark — modelled as `SimpleSubtypeSymbol` — or a type mark with a range. There is `integer` and there is `integer range 0 to 5`, but no `integer range`.

* `constraint` loses its `None` default and is typed `Range` instead of `Nullable[Range]`, in `ScalarConstraint` as well as in `ConstrainedScalarSubtypeSymbol`.
* The `Constraint` property returns `Range`.

## Documentation

* `ScalarConstraint` states why the range is mandatory, citing the `range_constraint` rule.
* `ConstrainedScalarSubtypeSymbol` gained a second example for the range-attribute form, `subtype index is natural range vector'range;`, which is a `RangeFromName` referring to a `RangeAttributeSymbol`. That form was the reason the parameter was optional, and it is a range like any other.

## Unit Tests

* `test_ScalarConstraint_RangeIsMandatory` and `test_ConstrainedScalarSubtypeSymbol_withoutConstraint` now assert that omitting the constraint raises a `TypeError`; previously both asserted that `Constraint` is `None`.
* `test_ScalarConstraint_WithRangeAttribute` covers the range-attribute form.

465 tests pass (was 464).

## Notes

`range <>` — as in `type matrix is array (natural range <>) of bit;` — is an `index_subtype_definition`, not a subtype indication. `ArrayType` models it as a dimension (`List[Range]`), so it never reaches this class.

**This is a breaking change**, so the next release is a minor one: v0.39.0 rather than v0.38.1. #168 has to be retitled and its version bump adjusted once this is merged.

**pyGHDL.dom needs a companion change.** `GetScalarConstrainedSubtypeFromNode` passes `None` when GHDL reports the range constraint as an `Attribute_Name`, with a `# TODO: Get actual range from AttributeName node?`. Nothing raises, so this is not a crash — but `Constraint` would hold `None` in violation of its type. `_Translate.py` already builds exactly the right object one function further down (`RangeFromName(node, RangeAttributeSymbol(node, GetName(node)))`), so the fix is a few lines. Filed in the findings notebook.

### DIFF
--- a/pyVHDLModel/Symbol.py
+++ b/pyVHDLModel/Symbol.py
@@ -873,10 +873,7 @@ class ScalarConstraint(Constraint, mixin=True):
 	"""
 	A mixin-class for a scalar constraint: a range.
 
-	The range is available as :data:`Constraint`. It is mandatory, because VHDL's ``range_constraint``
-	rule is ``range range``: a type mark is either written without a constraint at all, which is a
-	:class:`~pyVHDLModel.Symbol.SimpleSubtypeSymbol`, or with a range following the ``range`` keyword.
-	A form like ``integer range`` doesn't exist.
+	The range is available as :data:`Constraint`.
 
 	.. seealso::
 
@@ -897,7 +894,7 @@ class ScalarConstraint(Constraint, mixin=True):
 		"""
 		Read-only property to access the scalar type's range constraint (:attr:`_constraint`).
 
-		:returns: The constraint.
+		:returns: The constraint of the scalar subtype.
 		"""
 		return self._constraint
 
--- a/pyVHDLModel/Symbol.py
+++ b/pyVHDLModel/Symbol.py
@@ -873,28 +873,31 @@ class ScalarConstraint(Constraint, mixin=True):
 	"""
 	A mixin-class for a scalar constraint: a range.
 
-	The range is available as :data:`Constraint`.
+	The range is available as :data:`Constraint`. It is mandatory, because VHDL's ``range_constraint``
+	rule is ``range range``: a type mark is either written without a constraint at all, which is a
+	:class:`~pyVHDLModel.Symbol.SimpleSubtypeSymbol`, or with a range following the ``range`` keyword.
+	A form like ``integer range`` doesn't exist.
 
 	.. seealso::
 
 	   * :class:`Constrained scalar subtype symbol <pyVHDLModel.Symbol.ConstrainedScalarSubtypeSymbol>`
 	"""
-	_constraint: Nullable[Range]  #: The range constraining the scalar subtype, or ``None`` if unconstrained.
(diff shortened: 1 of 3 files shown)


pyVHDLModel/Symbol.py
```diff
@@ -873,28 +873,31 @@ class ScalarConstraint(Constraint, mixin=True):
 	"""
 	A mixin-class for a scalar constraint: a range.
 
-	The range is available as :data:`Constraint`.
+	The range is available as :data:`Constraint`. It is mandatory, because VHDL's ``range_constraint``
+	rule is ``range range``: a type mark is either written without a constraint at all, which is a
+	:class:`~pyVHDLModel.Symbol.SimpleSubtypeSymbol`, or with a range following the ``range`` keyword.
+	A form like ``integer range`` doesn't exist.
```
Paebbels:
```suggestion
	The range is available as :data:`Constraint`.
```

Paebbels:
CI: 53 jobs succeeded, 9 skipped, 1 failed — [run 30947957072](https://github.com/VHDL/pyVHDLModel/actions/runs/30947957072). All unit-test jobs pass on every platform and Python version.

The one failure is `PublishToGitHubPages / 📖 Merge multiple contents for publishing`, and it is not related to this change:

> Branch "claude/mandatory-scalar-constraint" is not allowed to deploy to github-pages due to environment protection rules.

The `github-pages` environment only accepts deployments from protected branches, so the job is rejected after one second, before any step runs. The same job failed identically on `claude/dependency-updates` ([run 30947167380](https://github.com/VHDL/pyVHDLModel/actions/runs/30947167380)), which is now merged as #171. Filed as a finding.

pyVHDLModel/Symbol.py
```diff
@@ -873,28 +873,31 @@ class ScalarConstraint(Constraint, mixin=True):
...
 		self._constraint = constraint
 
 	@readonly
-	def Constraint(self) -> Nullable[Range]:
+	def Constraint(self) -> Range:
 		"""
 		Read-only property to access the scalar type's range constraint (:attr:`_constraint`).
 
-		:returns: The constraint, or ``None`` if not set.
+		:returns: The constraint.
```
Paebbels:
```suggestion
		:returns: The constraint of the scalar subtype.
```

